PR #302, opened 2020-11-13
feature: Improve release notes index

pauloribeiro-codacy:
The page acting as the release notes index now includes links to the actual release notes pages.

For now, this page is updated manually, but in the future it may be possible to have the page update dynamically when we add new release notes.

Also, this page is a good place to let the users know that we intend to have regular Codacy Cloud release notes in the near future (or, put another way, that we aren't providing release notes for Cloud at the moment).

Here is a preview of the updated page:

https://github.com/codacy/docs/blob/feature/add-release-notes-index/docs/release-notes/index.md


pauloribeiro-codacy:
I'm merging this change now, but do let me know if you have feedback and we can quickly iterate on the solution. :+1:

PR #305, opened 2020-11-14
toolkit: Fix RSS feed dates

pauloribeiro-codacy:
Currently, the generated RSS feeds do not reflect the Git commit dates when the files where created or updated.

This only happens when the build or deploy happens on GitHub. Locally the dates are correct.

### DIFF
--- a/.github/workflows/mkdocs.yml
+++ b/.github/workflows/mkdocs.yml
@@ -16,9 +16,15 @@ jobs:
           fetch-depth: 0
       - run: |
           echo "Running git log"
+          git submodule init
+          git submodule update
           git submodule status
           echo "---"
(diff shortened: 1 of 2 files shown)


pauloribeiro-codacy:
I was able to identify the problem:

The MkDocs plugin [mkdocs-rss-plugin](https://github.com/Guts/mkdocs-rss-plugin/) fails to obtain the git log of files inside submodules using the latest version of Git ([2.29.2](https://github.com/git/git/releases/tag/v2.29.2)), which is the one used in the GitHub virtual environments.

I was testing locally with versions 2.17.1 and 2.25.1, but in this case, MkDocs could build the documentation successfully.

PR #314, opened 2020-11-17
clean: Move images

pauloribeiro-codacy:
Move images from the folder `/images` and closer to the Markdown source files that use them.

This will make it easier to manage and maintain the images and will make it possible to delete the folder `/images` once the previous documentation versions no longer use them (see https://github.com/codacy/docs/pull/308 and https://codacy.atlassian.net/browse/DOCS-133).


pauloribeiro-codacy:
It's better to merge this pull request only once all Codacy Self-hosted documentation versions are using their own versions of the files in `/images` instead of using the ones in `docs.codacy.com/images`.

PR #321, opened 2020-11-24
fix: Trigger Vale linter on push

pauloribeiro-codacy:


pauloribeiro-codacy:
Currently, Vale has an issue and always fails on the first commit of a pull request. So, we'll keep triggering Vale on `pull_request` and not on `push`.

PR #322, opened 2020-11-25
feature: Improve usability of version switcher with improved item labels

pauloribeiro-codacy:
We received feedback from some customers that have trouble figuring out which documentation version they should choose, namely if they're using Codacy Cloud.

This pull request tries to make the version switcher more user-friendly by explicitly mentioning the Codacy environment (Cloud or Self-hosted) that each version applies to:

image

### DIFF
--- a/.github/workflows/mkdocs.yml
+++ b/.github/workflows/mkdocs.yml
@@ -80,7 +80,7 @@ jobs:
       - name: Deploy docs
         run: |
           git fetch origin gh-pages --verbose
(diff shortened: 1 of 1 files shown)


ritamarquespires (dismissed):

pauloribeiro-codacy:
After "having slept on the issue", I think it makes more sense to have the label for the Latest version like this, listing the environment first and the "Latest" as a detail inside parenthesis:

image

We can always iterate again if we receive more feedback from customers.

PR #327, opened 2020-11-27
toolchain: Enable Codacy GitHub Action

pauloribeiro-codacy:
Followed the instructions provided by @franciscodua on this GitHub issue:

https://github.com/codacy/codacy-eslint/issues/920#issuecomment-733641578

### DIFF
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -1,0 +1,3 @@
+---
+exclude_paths: []
+
--- a/.github/workflows/codacy-analysis-cli.yml
+++ b/.github/workflows/codacy-analysis-cli.yml
@@ -6,7 +6,6 @@ jobs:
   codacy-analysis-cli:
     runs-on: ubuntu-latest
     name: Codacy Analysis CLI
-    if: github.ref != 'refs/heads/master' && !startsWith(github.ref, 'refs/heads/release/v')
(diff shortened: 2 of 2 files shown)


pauloribeiro-codacy:
Hi @franciscodua, I implemented the solution you described in your reply to the user.

I think I got it to work, but I have a few comments:

- I'm using the setting `max-allowed-issues: 2147483647` to allow the CLI to report a successful run even when it finds issues. Otherwise, it would fail the run even if the issues found are on files that weren't changed in the commit or pull request. :thinking: 
- Codacy will now wait for the CLI to run on each commit, however, I'm not 100% sure if Codacy isn't running the analysis again after receiving the results from the CLI.

pauloribeiro-codacy:
I disabled the Codacy setting **Run analysis through build server** on the codacy/docs repository because it was preventing the analysis from running on the other pull requests / branches.

pauloribeiro-codacy:
@machadoit confirmed that Codacy doesn't perform an analysis with the tools that already ran on the CLI.

So, with this solution, it's possible to run all tools configured for the repository on GitHub and then send the results to Codacy.

Then, with the GitHub integration enabled, Codacy reports issues as annotations on the modified lines.

PR #341, opened 2020-12-06
feature: Add minimum versions of self-hosted Git providers

pauloribeiro-codacy:
We must list the self-hosted Git provider versions that we support so that we can clearly communicate to customers these requirements while setting up Codacy Self-hosted, and also to help manage expectations when we must ask customers to upgrade their infrastructure. See [DOCS-38](https://codacy.atlassian.net/browse/DOCS-38) for more details.

Once this pull request is merged, we must also merge https://github.com/codacy/chart/pull/649 so that we mention the supported Git providers on the chart [system requirements](https://docs.codacy.com/chart/requirements/).

We should also have a process in place to make sure that we update these minimum versions whenever a change in Codacy Self-hosted requires that. This will also have the advantage of forcing us to make a conscious decision when we're introducing breaking changes to customers.

### DIFF
--- a/docs/faq/general/which-version-control-systems-do-you-support.md
+++ b/docs/faq/general/which-version-control-systems-do-you-support.md
@@ -13,6 +13,3 @@ Besides these, on [Codacy self-hosted](https://www.codacy.com/self-hosted) we al
 -   GitHub Enterprise (version 2.20.3 or later)
 -   GitLab Enterprise (version 12.6.2-ee or later)
 -   Bitbucket Server (version 6.6.0 or later)
-
-!!! note
-    We used to support the possibility to add other Git repositories to Codacy manually, but as part of a process to improve our performance, that option is no longer available. However, repositories that were added before the 12th Feb 2019 can continue to be analyzed if you configure [post-commit hooks](../../repositories/post-commit-hooks.md) on your Git repositories to notify Codacy of new commits.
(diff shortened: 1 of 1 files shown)


docs/faq/general/which-version-control-systems-do-you-support.md
```diff
@@ -10,9 +10,9 @@ On Codacy Cloud we support the following Git providers:
 
 Besides these, on [Codacy self-hosted](https://www.codacy.com/self-hosted) we also support:
 
--   GitHub Enterprise
--   GitLab Enterprise
--   Bitbucket Server
+-   GitHub Enterprise (version 2.20.3 or later)
+-   GitLab Enterprise (version 12.6.2-ee or later)
+-   Bitbucket Server (version 6.6.0 or later)
```
pauloribeiro-codacy, line 15: Which Git provider versions should we list here?

These are the versions of our Git provider instances for the dev environment, where I assume that we test the Codacy features against. But this doesn't necessarily mean that previous versions of the Git providers won't work with Codacy Self-hosted as well. :thinking:

pauloribeiro-codacy: For example, to address the issue [CY-3302](https://codacy.atlassian.net/browse/CY-3302?focusedCommentId=44559), we should list the minimum version **11.6** for GitLab Enterprise.

machadoit: We may want to make some specific effort on some specific version, and install that version instead, but I think that we should suggest the minimum as the version we have installed and test. So this would make sense for me for now

machadoit (dismissed):

pedrocodacy (approved):

pauloribeiro-codacy:
I added a note to clarify a bit better how some older versions of the self-hosted Git providers may still work with Codacy, but how Codacy does not make any guarantees regarding that and will not fix issues on those older versions:

image

davidscordeiro (approved):

PR #353, opened 2020-12-17
feature: Release notes for Codacy Self-hosted 3.2.0

pauloribeiro-codacy:
Details about the release on Slack: https://codacy.slack.com/archives/CSSSHAK9N/p1608131496003800

-  [x] Solve all TODO comments

### DIFF
--- a/docs/release-notes/self-hosted/self-hosted-v3.2.0.md
+++ b/docs/release-notes/self-hosted/self-hosted-v3.2.0.md
@@ -6,7 +6,7 @@ To upgrade Codacy, follow [these instructions](../../chart/maintenance/upgrade.m
 
 ## Product enhancements
 
--   Now, the list that allows adding organizations is paginated and supports displaying an indefinite number of organizations per Git provider. (CY-1938)
+-   Now, Codacy has an improved user experience to add and manage organizations and supports displaying an indefinite number of organizations per Git provider. (CY-1938)
 
 ## Bug fixes
 
--- a/docs/release-notes/self-hosted/self-hosted-v3.2.0.md
+++ b/docs/release-notes/self-hosted/self-hosted-v3.2.0.md
@@ -15,3 +15,45 @@ To upgrade Codacy, follow [these instructions](../../chart/maintenance/upgrade.m
 
 This version of Codacy Self-hosted includes the tool versions below. The tools that were updated on this version are highlighted in bold:
 
+-   Ameba 0.13.1
+-   Bandit 1.6.2
+-   Brakeman 4.3.1
+-   BundlerAudit 0.6.1
+-   Checkstyle 8.37
+-   Clang-tidy 10.0.1
+-   CodeNarc 1.6
+-   Coffeelint 2.1.0
+-   Cppcheck 2.2
+-   Credo 1.4.0
+-   CSSLint 1.0.5
+-   Detekt 1.14.2
+-   **ESLint 7.15.0 (updated from 7.14.0)**
+-   Faux-Pas 1.7.2
+-   Flawfinder 2.0.11
+-   Gosec v2.3.0
+-   Hadolint 1.18.2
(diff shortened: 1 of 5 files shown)


docs/release-notes/self-hosted/self-hosted-v3.2.0.md
```diff
@@ -0,0 +1,60 @@
+# Self-hosted v3.2.0
+
+These release notes are for [Codacy Self-hosted v3.2.0](https://github.com/codacy/chart/releases/tag/3.2.0){: target="_blank"}, released on December ##, 2020. <!-- TODO Update release date -->
+
+To upgrade Codacy, follow [these instructions](../../chart/maintenance/upgrade.md).
+
+## Product enhancements
+
+-   Now, Codacy has an improved user experience to add and manage organizations and supports displaying an indefinite number of organizations per Git provider. (CY-1938)
```
pauloribeiro-codacy, line 9: @inesbpinto @davidscordeiro, please check if you agree with the wording here. :+1:

docs/release-notes/self-hosted/self-hosted-v3.2.0.md
```diff
@@ -0,0 +1,57 @@
+# Self-hosted v3.2.0
+
+These release notes are for [Codacy Self-hosted v3.2.0](https://github.com/codacy/chart/releases/tag/3.2.0){: target="_blank"}, released on December ##, 2020. <!-- TODO Update release date -->
+
+To upgrade Codacy, follow [these instructions](../../chart/maintenance/upgrade.md).
+
+## Product enhancements
+
+-   To overcome timeouts when loading the repositories list page on Codacy Self-hosted, it's now possible to configure how many repositories to load at a time. To do this, set a value lower than the default of 100 by configuring `codacy-spa.config.codacy.pagination.repositoriesLimit` on your `values-production.yml` configuration file. (CY-3352)
```
pauloribeiro-codacy, line 9: @machadoit @jamescodacy, please check if the wording is appropriate.

I decided to include this in the "Product enhancements" because we ended up not actually fixing an issue, but instead adding more flexibility in the product to act as a workaround. But please let me know if you disagree.

machadoit: I like it! Maybe @heliocodacy can clarify on the `values-production.yml`, I think that the user usually has a values.yaml on their own right? What do we usually suggest to change on this cases?

I like the "To overcome timeouts" to avoid people messing with this value for free 😅

pauloribeiro-codacy: Please check the most recent version instead, I think it describes the new capability a bit better. :+1:

machadoit: I think I prefer your original proposal, and maybe mentioning the “takes a very long time to load” may be a mistake on my part. Since the aim is the user with the timeout.

My concern now, is that some people consider “long time to load” too soon, and change this. And this value is tricky, because lower values also have other problems, like calling the provider more times

davidscordeiro (approved):

inesbpinto (approved):

docs/release-notes/self-hosted/self-hosted-v3.2.0.md
heliocodacy: Whoever is installing Codacy, downloads a copy of the values-production.yaml file and updates what makes sense there. If there are breaking changes on this file, user will download the new one and update it with the values that make sense like encryption keys and connection strings.

pauloribeiro-codacy: Thanks for the extra context @heliocodacy! So this means that it makes sense to ask customers to change the new setting on their `values-production.yml` configuration file, right?

PR #414, opened 2021-01-20
feature: New "adding people" experience DOCS-113

pauloribeiro-codacy:
Refactors and updates the documentation to take into account the new "adding people" experience:

- [DOCS-113](https://codacy.atlassian.net/browse/DOCS-113)
- [CY-2643](https://codacy.atlassian.net/browse/CY-2643)

Preview of the main page covering the new feature:

https://github.com/codacy/docs/blob/feature/add-people-DOCS-113/docs/organizations/managing-people.md


docs/organizations/managing-people.md
```diff
@@ -0,0 +1,66 @@
...
+
+    **If you're using Bitbucket Cloud** the organization that contains your personal repositories can have other members since Bitbucket Cloud doesn't have a different type of [workspace](https://support.atlassian.com/bitbucket-cloud/docs/what-is-a-workspace/){: target="_blank"} for personal use.
+
+To list and manage the members of your organization open your organization **Settings**, page **People**:
+
+![People in an organization](images/organization-people.png)
+
+On this page you can see the list of members of your organization and when they last logged in to Codacy.
+
+<!-- TODO How to see if members are "regular members" or organization owners? -->
```
pauloribeiro-codacy, line 19: :construction: How to see if members are "regular members" or organization owners?

docs/organizations/managing-people.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Adding members to your organization
+
+Organization owners can add other team members to their organization on Codacy. This is useful to allow Codacy to analyze commits by contributors who haven't signed up for Codacy yet.
+
+To add members to your organization:
+
+1.  Open your organization **Settings**, page **People**, and click the button **Add people**.
+
+    <!-- TODO Consider adding another screenshot to show the UI path -->
```
pauloribeiro-codacy, line 46: :construction: Consider adding another screenshot to show how to navigate the UI.

docs/organizations/managing-people.md
```diff
@@ -0,0 +1,66 @@
+# Managing people
+
+Members of an organization can see the details of the repositories in that organization and add new repositories to be analyzed by Codacy. Besides this, **Codacy only analyzes commits to repositories from contributors who are members of the corresponding organization on Codacy**.
+
+!!! important
+    Make sure that your team members join your organization on Codacy so that Codacy analyzes their commits in the repositories of that organization.
+
+!!! note
+    The organizations on Codacy that contain your personal repositories can only have the owner of the repositories as their single member.
```
inesgomas, line 9: "Not true". You can add that email as author and it will be a billed seat. I would just remove this to avoid confusion. 

Personal orgs and normal orgs don't have any differences regarding seats(members) the "difference" that exists is that another member can not join your personal organization. But you can add emails do analyze freely and **it will update your people list accordingly.** 
I don't think this "difference" is worth mention. Specially because we will now not show any major visual difference between a "user" and "author" it would just confuse the user. It is just: People under this list are the ones being billed.

inesgomas: If you want to maintain the "Bitbucket Cloud" use case I understand and you can. But you could mention the real use case that is available only for BB which is "the personal organization showing up to them" not sure if a documentation about the People page is the right one.

docs/organizations/managing-people.md
inesgomas: You can’t we removed it due to technical limitations. People should rely on the roles they have on the provider side

docs/organizations/managing-people.md
```diff
@@ -0,0 +1,66 @@
...
+
+To list and manage the members of your organization open your organization **Settings**, page **People**:
+
+![People in an organization](images/organization-people.png)
+
+On this page you can see the list of members of your organization and when they last logged in to Codacy.
+
+<!-- TODO How to see if members are "regular members" or organization owners? -->
+
+## Joining an organization
```
inesgomas, line 21: Ok, here you could mention the BB special behaviour for personal organizations 👀

docs/organizations/managing-people.md
pauloribeiro-codacy: Thanks for the clarification @inesgomas, and your suggestions do make sense: it's better to just mention the "positive" exception that users can join someone else's personal organizations on Bitbucket. :+1:

pauloribeiro-codacy: Thinking a bit more about this, I'm considering not even mentioning the Bitbucket "exception" at all, since on this Git provider there's not really the concept of "personal repository". Instead, all repositories must belong to a Workspace:

image

These Workspaces map to Codacy organizations, but once again there's no real mapping to the concept of "personal organization":

image

I think that until we have more information that tells us otherwise, it's simpler for the users if we omit this detail. :+1:

inesgomas: That is also true 😄 Maybe not mentioning is the best

inesgomas (approved):

docs/organizations/managing-people.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Joining an organization {: id="joining"}
+
+To become a member of an organization on Codacy you must [sign up to Codacy](../getting-started/getting-started-with-codacy.md) using your Git provider and follow the instructions to either join an existing organization or add a new one.
+
+To join or add an organization after having complete the signup process, click **Organizations** on the top right-hand menu under your avatar:
+
+![Joining an organization](images/organization-join.png)
+
+Organization owners can control who joins their organization by choosing one of the following options on the page **Plan and Billing** of the organization:
```
inesgomas, line 20: This is also only for cloud

pauloribeiro-codacy: Good catch!

inesgomas (dismissed):

alerizzo (approved):

PR #445, opened 2021-02-09
fix: ESLint doesn't analyse JSON anymore

lolgab:


pauloribeiro-codacy (approved):
Thanks for the update! :rocket:

PR #456, opened 2021-02-12
clean: Improve UI copy for "Accepting new people to Codacy" DOCS-212

Gsandec:
By showing "Request access to join" only, I believe it immediately communicates what's different from the first one. Adding "New people need to" at the beginning of the sentence makes it hard to understand for fast readers.


pauloribeiro-codacy:
> By showing "Request access to join" only, I believe it immediately communicates what's different from the first one. Adding "New people need to" at the beginning of the sentence makes it hard to understand for fast readers.

On the documentation we replicate the exact labels that appear on the UI when referring to them so that it's easier for users to "make the connection" and also to be confident that the feature is the same.

However, you are right that the current copy could be improved. My suggestion is that we not only update the documentation, but also update the UI copy as well. For example, something like this rough proposal (in fact, I would like to spend some time improving the copy of the whole section and not just that option):

image

@inesgomas @alerizzo, do you think it's OK to tweak this? Can I make a proposal and involve you in the review? :muscle:

inesgomas:
@pauloribeiro-codacy I got really confused with "approve request to join" suggestion, I think the copy there is more "What the other users can do" not what "you can do"
If you want to go there, than the other options need to be changed as well

pauloribeiro-codacy:
> @pauloribeiro-codacy I got really confused with "approve request to join" suggestion, I think the copy there is more "What the other users can do" not what "you can do"

Yes, I did a really rough proposal but my rationale was that this copy is meant for the organization owners (the only ones able to change this setting). So I agree with you, the copy must clarify "who does what". :slightly_smiling_face: 

> If you want to go there, than the other options need to be changed as well

I will make a full proposal when I have some time to think more carefully about it. :+1:

pauloribeiro-codacy:
@Gsandec and @inesgomas, I made a proposal for improving the copy of these configurations. Please have a look:

https://docs.google.com/document/d/1YcnQTtBIGKG2U8gTOiEthu_TmdoBkDFYRxAYT2R7UlA/edit#heading=h.4m9a8hczc5tt

pauloribeiro-codacy:
We made a few adjustments both to the UI copy on the Codacy UI and to the documentation. Here's how it looks now:

image

@Gsandec please have a look. :+1:

inesgomas (dismissed):

inesgomas:
I am not a fan of that screenshot tho. Empty page with only this setting? How did you came up with that state? 😅

pauloribeiro-codacy:
> I am not a fan of that screenshot tho. Empty page with only this setting? How did you came up with that state?

You're right, I did some "editing" to simplify because I thought that the plan / billing / seats information on top could be distracting. :slightly_smiling_face: Do you think the following screenshot is "vanilla" enough not to be distracting because customers have different info? We can use this one instead. :+1: 

image

inesgomas (approved):

PR #491, opened 2021-02-22
fix: Remove add badge reference in repository dashboard documentation CY-3774

inesgomas:
In the meantime while we are testing out tours the add badge was removed. So I am updating the documentation accordingly. 
In a short future we should have here "Start tour" button


pauloribeiro-codacy (approved):
This is great @inesgomas. :+1: :rocket:

I also removed the button from the full screenshot (that is used both on this page and also on the [Getting started with Codacy](https://docs.codacy.com/getting-started/getting-started-with-codacy/#add-repositories)).

PR #501, opened 2021-02-25
fix: Remove spurious Gosec update

pauloribeiro-codacy:
The false positive was caused by the removal of the initial "v" in the version number:

https://github.com/codacy/codacy-gosec/commit/cecba6d73e0a26e15ff3246a6cd58962ea41990f


pauloribeiro-codacy:
Thank you @ameliavp for noticing this! :tada:

PR #520, opened 2021-03-03
feature: Update configure ignore issues permissions doc CY-3647

franciscodua:


pauloribeiro-codacy (approved):
These updates where we remove limitations or exceptions for Git providers are just great! :star_struck: :raised_hands:

pauloribeiro-codacy:
I changed the title of the pull request to include `feature:` instead of `doc:`, since otherwise, almost all updates to this repository would have the same prefix. :slightly_smiling_face:

pauloribeiro-codacy (approved):

pauloribeiro-codacy (approved):

pauloribeiro-codacy (changes requested):
I remembered that we should also update the screenshot since the copy of the **Member privileges** screen will change according to [these mockups](https://www.figma.com/file/qJth2Cn2aYZ9a4mMGdNuDt/Flexibility-over-member-privileges?node-id=1%3A171).

franciscodua:
>I remembered that we should also update the screenshot since the copy of the Member privileges screen will change according to these mockups.

We didn't change the page yet. That will happen in future epics

pauloribeiro-codacy:
> We didn't change the page yet. That will happen in future epics

Thanks for the heads up @franciscodua, it makes sense. We will have an opportunity to revisit this documentation and update the screenshot then. :+1:

PR #526, opened 2021-03-05
fix: Rework how to "set default" and "reset" the tool patterns configuration

pauloribeiro-codacy:
Uses the information collected by @inesgomas on [how the default patterns work](https://whimsical.com/default-patterns-flow-3FcZQCMpiorKUdjmR5xY4H) to fix an incorrect instruction on the **Code patterns** page.

Before, the instructions assumed that the **Reset** operation on the menu would change the pattern configuration for the tool to "defaults provided by Codacy", but in fact, the reset loads the account default settings instead.


docs/repositories-configure/code-patterns.md
```diff
@@ -50,19 +50,13 @@ To apply the same pattern configuration to multiple repositories:
...
+After defining the default pattern configuration for your account, the new repositories that you add will automatically use those pattern configurations.
 
-    The new repositories will automatically use the tool pattern configuration that you set as default.
-
-To reset the pattern configuration of a tool back to the default provided by Codacy:
-
-1.  Expand the menu at the top of the pattern list for that tool and select **Reset** to revert the configuration back to the one provided by Codacy.
-
-1.  Select **Set default** on the same menu to reset the default configuration.
+Alternatively, you can also manually update the pattern configuration of a tool to use your account configurations by expanding the menu at the top of the pattern list and selecting **Reset**.
```
inesgomas, line 59: @machadoit when we use `Set Default/Reset` is it by the tool you are currently on or is it all the patterns of the repository?

pauloribeiro-codacy: Interesting question @inesgomas. :thinking: Since the action is on a menu associated with a tool, I was expecting that it would only affect that tool. But let's wait for the reply from @machadoit.

inesgomas: I don't want to make that assumption just because of the positioning, codacy has made decisions like this before

pauloribeiro-codacy: I'll do a quick test just to confirm and adapt the documentation. I'll let you know about the changes.

@inesgomas, do you think this deserves a clearer UI (perhaps even an updated copy could help). I can create a task for this.

pauloribeiro-codacy: I just tested and indeed the operations apply to all tools, so I made this change: https://github.com/codacy/docs/pull/526/commits/4b517dcf91b60b607b4761cf9caa1fb10589197a

I'm going to propose an updated copy for the confirmation dialog window that will help call attention to this.

PR #535, opened 2021-03-08
fix: Remove error message for line ending CY-3746

franciscodua:


josemiguelmelo (dismissed):

pauloribeiro-codacy (commented):
This page (and the error message) was created together with [CY-2799](https://codacy.atlassian.net/browse/CY-2799).

At the time, the error message was introduced on https://bitbucket.org/%7B%7D/%7Ba820bc98-1bd2-411e-86ed-05c9e08d5235%7D/pull-requests/1837 but was later reverted on https://bitbucket.org/%7B%7D/%7Ba820bc98-1bd2-411e-86ed-05c9e08d5235%7D/pull-requests/1917...

@pedrocodacy do you know why we ended up not adding the error message? It would make it much easier for users to troubleshoot the issue and it's probably the only way they could find the documentation page to be able to solve the issue.

pedrocodacy (approved):

pauloribeiro-codacy (approved):
Let's sync the docs with the product, but if this issue comes up frequently we should consider presenting an explicit error message to the users.

PR #551, opened 2021-03-16
 feature: Mention SwiftLint support for reporting complexity issues DOCS-232

pauloribeiro-codacy:
SwiftLint includes a rule that reports issues on complex functions:

https://realm.github.io/SwiftLint/cyclomatic_complexity.html

We decided to include SwiftLint under the column **Complexity** on the page [Supported languages and tools](https://docs.codacy.com/getting-started/supported-languages-and-tools/) with a note clarifying this (limited) support for Swift complexity on Codacy.

image

### DIFF
--- a/docs/getting-started/supported-languages-and-tools.md
+++ b/docs/getting-started/supported-languages-and-tools.md
@@ -43,7 +43,7 @@ The table below lists all programming languages that Codacy supports and the cor
     </tr>
     <tr>
       <td>C</td>
-      <td><a href="https://clang.llvm.org/extra/clang-tidy/">Clang-Tidy</a><a href="#client-side"><sup>*</sup></a>,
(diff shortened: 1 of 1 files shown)


peloycosta:
I agree with the text, but I would also add specifically what it supports, which is setting thresholds to get issues for the complexity. To make it clear that we do not give any complexity values. Not sure how to phrase it though.

davidscordeiro (dismissed):

pauloribeiro-codacy:
> I agree with the text, but I would also add specifically what it supports, which is setting thresholds to get issues for the complexity. To make it clear that we do not give any complexity values. Not sure how to phrase it though.

I see what you mean :+1:, please see the new version.

peloycosta (approved):

PR #560, opened 2021-03-22
feature: List frameworks supported by Checkov CY-4038

pauloribeiro-codacy:
As a first improvement to make it clearer that Codacy supports a set of infrastructure-as-code frameworks, we can list the frameworks directly in the column **Languages** of the [Supported languages and tools](https://docs.codacy.com/getting-started/supported-languages-and-tools/).


davidscordeiro (approved):
awesome!

PR #563, opened 2021-03-24
fix: replace screenshot for ignored issues permissions

pedrocodacy:
Updating this screenshot since we are releasing this feature officially today and this will not match our UI changes


pauloribeiro-codacy (approved):
Makes sense! :shipit:

PR #571, opened 2021-03-26
fix: Include tools in configure patterns permissions table

josemiguelmelo:


pedrocodacy (approved):

pedrocodacy:
we decided to propose this change since the feature we are introducing also manages permissions to change what tools are on/off for a repo

pauloribeiro-codacy (approved):
Although I would assume that "Configuring patterns" implicitly also includes toggling tools, this is a nice clarification. :+1:

PR #580, opened 2021-04-01
feature: Update file extensions permissions information CY-4018

franciscodua:


docs/organizations/roles-and-permissions-for-synced-organizations.md
```diff
@@ -16,6 +16,7 @@ Depending on your role on the Git provider you will have different permissions o
       <th>View repository</th>
       <th>Invite and accept members</th>
       <th>Modify billing</th>
+      <th>Configure file extensions</th>
```
franciscodua, line 19: Maybe we should think if we want to keep adding columns or if we want to make this a bit more generic (example: Configure analysis)

pedrocodacy: You are very right sir, I just looked at the current page in my 13'' monitor and I think this either wont fit or it will look very cluttered.

Can we add this to the "Configure tools and patterns" column ("Configure tools, patterns and file extensions")?

franciscodua: https://github.com/codacy/docs/pull/583
I created this PR with this suggestion

josemiguelmelo (approved):

pedrocodacy (approved):

PR #599, opened 2021-04-13
toolchain: Force remark-lint to check for atx style headings

prcr:
On pages that include a YAML front matter remark-lint was suggesting the [setext syntax](http://webcoder.info/markdown-headers.html#the-case-for-setext-headers), probably to be consistent with the separators of the YAML block (---).


prcr:
Used the solution in https://github.com/remarkjs/remark-lint/issues/211 as reference.

PR #677, opened 2021-05-17
feature: Using pagination when calling API endpoints DOCS-245

prcr:
Adds a section to the Codacy API overview with instructions and an example on how to call endpoints that use pagination.

This page should now mention all the basic mechanics of how to use the Codacy API so that the "end-to-end examples" that I'm going to work on next don't have to explain the basics.

Preview of the new section:

https://github.com/codacy/docs/blob/feature/api-overview-pagination-DOCS-245/docs/codacy-api/using-the-codacy-api.md#using-pagination

### DIFF
--- a/docs/codacy-api/using-the-codacy-api.md
+++ b/docs/codacy-api/using-the-codacy-api.md
@@ -108,7 +108,7 @@ Endpoints that return lists containing a potential large number of results use c
 -   Calling endpoints that use pagination returns the first page of results together with a `pagination` object that includes a `cursor`.
 -   To obtain the next page of results, call the endpoint again using the `cursor` from the previous response as a parameter.
 -   If the response doesn't include a `cursor` it means that the endpoint returned the last page of results.
--   Use the parameter `limit` to configure the number of results that the endpoint returns in each page. The maximum value for `limit` is 1000.
+-   Use the parameter `limit` to configure the number of results that the endpoint returns in each page. The maximum `limit` is 1000 and the default is 100.
 
 !!! note
     To make sure that you receive all results when calling an endpoint with pagination, repeat the process above until the response doesn't include the cursor to obtain another page of results.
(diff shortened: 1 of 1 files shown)


prcr:
@alerizzo @inesgomas unfortunately, I totally forgot to ask for your review on the previous section that I added about how to pass parameters to endpoints. If you have the time I would appreciate it if you could also do a quick read of that section and let me know if I should open a separate pull request to work on that. :+1: :bow:

https://github.com/codacy/docs/blob/feature/api-overview-pagination-DOCS-245/docs/codacy-api/using-the-codacy-api.md#using-parameters-in-requests

docs/codacy-api/using-the-codacy-api.md
```diff
@@ -95,8 +95,63 @@ curl -X GET https://app.codacy.com/api/v3/analysis/organizations/gh/codacy/repos
...
      -H 'api-token: SjE9y7ekgKdpaCofsAhd' \
      -H 'Content-Type: application/json' \
      -d '{"levels": ["Error", "Warning"]}'
 ```
+
+## Using pagination
+
+Endpoints that return lists containing a potential large number of results use cursor-based pagination to return the results in small batches:
+
+-   Calling endpoints that use pagination returns the first page of results together with a `pagination` object that includes a `cursor`.
```
inesgomas, line 108: I was very confused with the beginning of the sentence, I suggest something like this

```suggestion
-   These endpoints will return the results together with a `pagination` object that includes a `cursor`.
```

inesgomas: I would avoid saying "first page" because it is a condition for all the pages

madalenamelo (dismissed):
I agree with @inesgomas 's feedback, I think her suggestion makes it a bit simpler and clearer. Other than that, these changes answer most of the questions we get (via Intercom at least) about the API

docs/codacy-api/using-the-codacy-api.md
prcr: Good feedback, initially I had these explanations intermixed with the examples so they would tell a "story". But now it makes more sense to have this more generic. :+1:

inesgomas (approved):

PR #684, opened 2021-05-21
feature: Update minimum supported GitLab Self-managed version

prcr:
Extra context on the following Slack thread:
https://codacy.slack.com/archives/C8X9SS1H7/p1621565381017700


prcr:
We're not updating the minimum supported version for GitLab Self-managed at this time.

PR #715, opened 2021-06-14
fix: Update short tool names in "Codacy configuration file"

prcr:
Updates the list of tool short names in the page [Codacy configuration file](https://docs.codacy.com/repositories-configure/codacy-configuration-file/#which-tools-can-be-configured-and-which-name-should-i-use) with the updated values retrieved using the API endpoint `https://api.codacy.com/api/v3/tools`.


tools/README.md
```diff
@@ -20,3 +20,20 @@ optional arguments:
   -h, --help     show this help message and exit
   -v, --verbose  False
 ```
+
+## list-tool-short-names.sh
+
+Outputs the list of short names for all supported tools as returned by the Codacy API endpoint [tools](https://api.codacy.com/api/v3/tools). Useful to update and synchronize the list of tools in [Which tools can be configured and which name should I use?](https://docs.codacy.com/repositories-configure/codacy-configuration-file/#which-tools-can-be-configured-and-which-name-should-i-use).
```
prcr, line 26: If we validate that the configuration file works for all supported tools, later we can use this automation to either:

- Build the list of tools automatically during the deployment of the documentation, or
- Periodically check if there are any updates to the list.

lolgab: There are some settings in the codacy configuration file that aren't dependent on the tool.. and are set in the platform before running the tool itself. So I'd say it is correct to add all the tools here.
The only question I have is about the manual client side tools like clang-tidy where for sure the codacy config file is not used..

prcr: That's a good point @lolgab! I adapted the script and updated the list to exclude client-side tools. :+1:

prcr:
Confirmed with @lolgab that the results from the API are the most up-to-date and can be considered the "source of truth". raised_hands

PR #814, opened 2021-08-31
test: Trigger a Vale issue

prcr:
Quick test to check if the Vale GitHub action correctly reports issues only on lines that were modified in each pull request.


prcr:
Vale seems to be working as intended. :+1:

PR #918, opened 2021-11-08
feature: Add Spectral to the tools supported

vascorsd:
We are adding support the [Spectral](https://stoplight.io/open-source/spectral/) tool which is a generic linter for yaml and json.

At this point we are not supporting user defined configuration files and only our rules.

The rules we included in the system by default are only able to validate [OpenAPI](https://www.openapis.org/) and [AsyncAPI](https://www.asyncapi.com/).

For the table of the supported languages I added sections for those (OpenApi and AsyncAPI) but at the product level when it appears in the Code Patterns we don't have a way to talk about those in specific and the tool appears as supported by Json and Yaml.

So the question here is, should we proceed like this and add these new entries or should approach this differently? It seems we already have similar situation for the CloudFormation and other tools, so maybe it would be a good idea to point out the situation somehow in the docs.

@prcr @pedrocodacy @lolgab

### DIFF
--- a/docs/getting-started/supported-languages-and-tools.md
+++ b/docs/getting-started/supported-languages-and-tools.md
@@ -46,6 +46,13 @@ The table below lists all languages and frameworks that Codacy supports and the 
       <td>-</td>
     </tr>
     <tr>
+      <td>AsyncAPI</td>
+      <td><a href="https://stoplight.io/open-source/spectral/">Spectral</a></td>
+      <td></td>
+      <td>-</td>
+      <td>-</td>
+    </tr>
+    <tr>
       <td>AWS Cloud​Formation</td>
       <td><a href="https://github.com/bridgecrewio/checkov/">Checkov</a></td>
       <td></td>
@@ -201,6 +208,13 @@ The table below lists all languages and frameworks that Codacy supports and the 
     <tr>
       <td>Objective-C</td>
       <td><a href="https://clang.llvm.org/extra/clang-tidy/">Clang-Tidy</a><a href="#client-side"><sup>1</sup></a>, <a href="http://fauxpasapp.com/">Faux Pas</a><a href="#client-side"><sup>1</sup></a></td>
+      <td></td>
+      <td>-</td>
+      <td>-</td>
+    </tr>
+    <tr>
+      <td>OpenAPI</td>
+      <td><a href="https://stoplight.io/open-source/spectral/">Spectral</a></td>
       <td></td>
       <td>-</td>
       <td>-</td>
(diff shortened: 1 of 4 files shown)


prcr:
Since we're not supporting custom configuration files for Spectral, we must also mention this tool in the list of tools that don't support configuration files:

https://docs.codacy.com/repositories-configure/configuring-code-patterns/#using-your-own-tool-configuration-files

Finally, we should also include the "[short name](https://github.com/codacy/codacy-spectral/blob/master/src/toolMetadata.ts#L1)" of this tool in the [Codacy configuration file page](https://docs.codacy.com/repositories-configure/codacy-configuration-file/#which-tools-can-be-configured-and-which-name-should-i-use).

(Unfortunately, the information about the supported tools is scattered throughout the documentation, which makes it harder to add all relevant information! I have already identified this and there's [an issue](https://codacy.atlassian.net/browse/DOCS-165) to think about improving this, though. :+1:)

prcr (changes requested):
Since we're not supporting custom configuration files for Spectral, we must also mention this tool in the list of tools that don't support configuration files:

https://docs.codacy.com/repositories-configure/configuring-code-patterns/#using-your-own-tool-configuration-files

Finally, we should also include the "[short name](https://github.com/codacy/codacy-spectral/blob/master/src/toolMetadata.ts#L1)" of this tool in the [Codacy configuration file page](https://docs.codacy.com/repositories-configure/codacy-configuration-file/#which-tools-can-be-configured-and-which-name-should-i-use).

(Unfortunately, the information about the supported tools is scattered throughout the documentation, which makes it harder to add all relevant information! I have already identified this and there's [an issue](https://codacy.atlassian.net/browse/DOCS-165) to think about improving this, though. :+1:)

prcr:
> So the question here is, should we proceed like this and add these new entries or should approach this differently? It seems we already have similar situation for the CloudFormation and other tools, so maybe it would be a good idea to point out the situation somehow in the docs.

You're right, Spectral is very similar to Checkov in the way it validates the semantics of particular specification formats and not just the basic JSON/YAML syntax. For now, I'm happy to leave as-is:

-   The page "[Supported languages and tools](https://docs.codacy.com/getting-started/supported-languages-and-tools/)" is targeted more to prospects evaluating Codacy so it's important that we mention the actual languages/specifications/frameworks that our tools are able to validate.
-   On the other hand, the person configuring the code patterns will be a DevOps Engineer or Administrator who is configuring these settings for their teams. In principle, the person making this configuration knows that the OpenAPI specification consists of a JSON or YAML file, for example, and shouldn't be surprised to find the tool listed under these "languages". :+1:

vascorsd:
Thanks for the guidance. I think I changed all the pages mentioned.

prcr (approved):
I pushed another commit just to sort the list of tools on the "Configuring code patterns page" so it looks nice and tidy. :nerd_face: 

Looks good to me! :shipit:

vascorsd:
> I pushed another commit just to sort the list of tools on the "Configuring code patterns page" so it looks nice and tidy. nerd_face
> 
> Looks good to me! :shipit:

I was thinking the same, but didn't want to change too much 😅.

PR #927, opened 2021-11-18
draft: List tools that support parameter configuration

mrfyda:
This is a quick list to check which tools support parameter configuration. I'm not sure what page to add this.
I couldn't find any mention, so perhaps we could also add an explanation of what is parameter configuration.

@prcr if you give me some guidance, I'm happy to take this

### DIFF
--- a/docs/repositories-configure/configuring-code-patterns.md
+++ b/docs/repositories-configure/configuring-code-patterns.md
@@ -66,8 +66,8 @@ For example, when adding a new repository on Codacy you can copy the tool and co
 
     The following example illustrates the points above:
 
-    | Source repository | Target repository | Target repository after import |
-    |-------------------|-------------------|--------------------------------|
+    | Source repository                                                    | Target repository                                                    | Target repository after import                                                   |
+    | -------------------------------------------------------------------- | -------------------------------------------------------------------- | -------------------------------------------------------------------------------- |
     | ![Source repository](images/code-patterns-import-example-source.svg) | ![Target repository](images/code-patterns-import-example-target.svg) | ![Target repository after import](images/code-patterns-import-example-final.svg) |
 <!--code-patterns-copy-end-->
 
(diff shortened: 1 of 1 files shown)


prcr:
> This is a quick list to check which tools support parameter configuration. I'm not sure what page to add this. I couldn't find any mention, so perhaps we could also add an explanation of what is parameter configuration.
> 
> @prcr if you give me some guidance, I'm happy to take this

Our current information architecture for the information on supported tools isn't the best one since the information is scattered across several pages (see [DOCS-165](https://codacy.atlassian.net/browse/DOCS-165)):

- The page where you added the new information is one of the few places where we list all supported tools; however, that section is dedicated to listing the tools that support custom configuration files so it would be confusing to also include there which tools support parametrization of rules on the Codacy UI. :thinking: 
- The only other page where we have a direct list of tools is the [Codacy plugin tools](https://docs.codacy.com/related-tools/codacy-plugin-tools/). Perhaps we could include the new information there, and have a link from the page [Configuring code patterns](https://docs.codacy.com/repositories-configure/configuring-code-patterns/) to this page.
- However, this would contribute to the issue of having the information scattered across too many pages and is not ideal.

Before investing more time in this, I would like to know if it's generally useful for customers to know this information. [I've asked Harry](https://codacy.slack.com/archives/C8X9SS1H7/p1637341856011200?thread_ts=1637238926.007400&cid=C8X9SS1H7) for more details on the use case of the customer who wanted to know this information.

prcr:
@mrfyda after [Harry's reply](https://codacy.slack.com/archives/C8X9SS1H7/p1637580572012200?thread_ts=1637238926.007400&cid=C8X9SS1H7) I'm a bit unsure if this information could be useful for other customers or not, since it seems quite a specific request. :thinking: 

If you consider it worthwhile to make the information of which tools support configurable rules, I'd say that we can add it to the [Codacy plugin tools page](https://docs.codacy.com/related-tools/codacy-plugin-tools/) almost as a placeholder so I don't forget to include this information somewhere while tackling [DOCS-165](https://codacy.atlassian.net/browse/DOCS-165). What do you think?

PR #951, opened 2021-12-14
clean: Remove section on how to set/reset default code patterns CY-5032

prcr:
The option to set and reset default tool and code pattern configurations is no longer available on Codacy.


inesgomas (approved):
🎉

PR #1018, opened 2022-02-07
feature: Exclude release notes from search results

prcr:
**Excludes the release notes pages** from the search index to improve the quality of the search results in some situations, such as when searching for the name of a tool:

image

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -81,7 +81,6 @@ plugins:
           exclude:
               - release-notes/cloud/*
               - release-notes/self-hosted/*
-              - special-thanks.md
     # Include the last modified date at the bottom of each page
     # https://github.com/timvink/mkdocs-git-revision-date-localized-plugin
     - git-revision-date-localized
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -75,6 +75,13 @@ plugins:
     # https://github.com/apenwarr/mkdocs-exclude
     - exclude:
           glob: "assets/includes/*.md"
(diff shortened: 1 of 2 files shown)


prcr:
@IGFCoimbra and @madalenamelo, what do you think about this change? Is it useful, or do you sometimes need to find information on the release notes pages? :thinking:

stefanvacareanu7:
I don't think it's going to make a difference for ourselves.
For the customers, I may be afraid if they are trying to search for something, notice there is a release page, and get no results, to assume there's no info for them.
So, from my point of view, taking out the feature of searching in the release notes for customers without giving another such option is not necessarily a good idea.

madalenamelo (approved):
I understand Stefan's concern but I think removing the release notes from the search results will give more visibility to results that are likely more relevant. And the release notes are still available on the side bar and can be easily found so I think this change makes sense

prcr:
> Why are we removing the special thanks page alongside the release notes?

Fair question, I got the idea of also excluding that page because it includes the names of several tools (so it can often come up in search results), but since it's so outdated it's not really useful at the moment.

Since you brought this up, I now think it's better and "cleaner" to just exclude the release notes for now, and think about either updating or removing the "Special thanks" page as part of a [different task](https://codacy.atlassian.net/browse/DOCS-165?focusedCommentId=48233). :+1:

prcr:
I'll apply this change, but we can always revert it if we notice that customers need to use the search to find information in the release notes. @stefanvacareanu7

PR #1045, opened 2022-02-16
feature: Add guide to use API to import list of emails to an organization

mrfyda:
idea from https://codacy.productboard.com/insights/notes/all-notes/notes/21744885
would be interesting to understand if users find this useful


mrfyda:
seems to include some unintended submodules updates 😮 oops

prcr:
> seems to include some unintended submodules updates open_mouth oops

No worries, I pushed another commit to revert those changes. :+1:

prcr:
Amazing contribution, @mrfyda! :raised_hands: :tada: 

Everything seems perfect, except I'm having trouble understanding the rationale for the expected CSV format. :thinking: Besides this, it seems that jq includes extra `"\n"` values in the output array:

```bash
$ cat emails.csv  | jq --raw-input --raw-output --compact-output --slurp 'split(",") | map(rtrimstr("\r\n") | select(length > 0)) | .[1:]'
["\n","no-reply@codacy.com","\n"]
```

I can have a better look at this later.

prcr:
@mrfyda I was thinking about this, and I think it could be preferable to simplify the example script around the API call so it's easier to understand by users. For that, we could accept a simpler file containing one email address per line instead of a CSV file. Developers can always obtain this simpler file from a CSV file containing the email addresses if needed.

So, if the input files contained one email address per line:

```bash
$ cat emails.txt
email1@codacy.com
email2@codacy.com
email3@codacy.com

email4@codacy.com
```

The jq command could be more straightforward (I also removed the flag `--compact-output` since it doesn't really affect the behavior of the API call):

```bash
$ cat emails.txt | jq --raw-input '[inputs | select(length>0)]'
[
  "email2@codacy.com",
  "email3@codacy.com",
  "email4@codacy.com"
]
```

Everything else is good to go, I'm just concerned about having questions from users about the jq command instead of the actual API endpoint. :nerd_face:

mrfyda:
thank you for the feedback! 🙌 
updated!

docs/codacy-api/examples/adding-people-to-codacy-programmatically.md
```diff
@@ -0,0 +1,72 @@
...
+The example script:
+
+1.  Defines the [account API token](../api-tokens.md#account-api-tokens) used to authenticate on the Codacy API.
+1.  Defines the path and filename of the file containing the email addresses list.
+1.  Uses [jq](https://github.com/stedolan/jq){: target="_blank"} to read the email addresses list from a file.
+1.  Calls the Codacy API endpoint [addPeopleToOrganization](https://app.codacy.com/api/api-docs#addpeopletoorganization){: target="_blank"} to add a list of email addresses to Codacy.
+
+```bash
+CODACY_API_TOKEN="<your account API token>"
+FILENAME="emails.text"
```
prcr, line 49: Just one missing detail! I'm pushing this change and merging. :rocket: 

```suggestion
FILENAME="emails.txt"
```

prcr (approved):
Well done! :shipit: :rocket:

mrfyda:
> Everything else is good to go, I'm just concerned about having questions from users about the jq command instead of the actual API endpoint. 🤓

using a simple text file, it should be possible to avoid using `jq` all together
do you mind if we look at that in a second iteration?

docs/codacy-api/examples/adding-people-to-codacy-programmatically.md
mrfyda: 🤦

prcr:
> > Everything else is good to go, I'm just concerned about having questions from users about the jq command instead of the actual API endpoint. nerd_face
> 
> using a simple text file, it should be possible to avoid using `jq` all together do you mind if we look at that in a second iteration?

Sure! The idea is that we should try to make working/useful examples with as little unnecessary complexity as possible. They're just a starting point for developers to adapt to their specific use cases. :+1:

PR #1056, opened 2022-02-18
feature: Add links to Adding people to Codacy programmatically

prcr:
We can unobtrusively highlight the new use case of adding people programmatically using "See also" links and the admonition card on the "Managing people" page. :bulb:

### DIFF
--- a/docs/codacy-api/examples/adding-people-to-codacy-programmatically.md
+++ b/docs/codacy-api/examples/adding-people-to-codacy-programmatically.md
@@ -69,4 +69,4 @@ email4@codacy.com
 
 ## See also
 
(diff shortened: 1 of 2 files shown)


docs/codacy-api/examples/adding-people-to-codacy-programmatically.md
```diff
@@ -69,4 +69,4 @@ email4@codacy.com
 
 ## See also
 
--   [API tokens](../api-tokens.md)
```
prcr, line 72: The other example page had a "See also" link to the API tokens because it was about creating API tokens programmatically. But here we can link to "Adding people instead".

mrfyda (approved):

PR #1066, opened 2022-02-22
toolchain: Check that all pull request tasks are complete

prcr:
Uses the following GitHub action to check if all tasks in a pull request are complete:

https://github.com/marketplace/actions/task-completed-checker

This is useful to block merging pull requests where a few steps are still missing.

- [ ] This is a test


prcr:
I'm closing this pull request because I don't like the idea of having the new check fail the CI pipeline, as it will mask useful information from the status of the other workflows.

PR #1026, opened 2022-02-11
feature: Mention drill down on overall quality chart CY-5608

prcr:
Updates the Organization Dashboard documentation to mention that it's possible to click the bars to see the list of grouped repositories.

### :eyes: Live preview
https://deploy-preview-1026--docs-codacy.netlify.app/organizations/organization-overview/#overall-quality-chart

PR #1079, opened 2022-02-28
clean: Control beta feature visibility with environment variable CY-5608

prcr:
Adds a "feature flag" to help control the visibility of Beta features on the documentation. When deploying the documentation for Self-hosted versions, the flag is set to `false`, thus hiding all Beta features on the Self-hosted documentation.

This pull request also places the changes from #1026 behind the new flag, as per [this decision](https://codacy.slack.com/archives/CA8A4L1DY/p1645788481492139?thread_ts=1645710065.068459&cid=CA8A4L1DY).

### DIFF
--- a/.github/styles/Vocab/Codacy/accept.txt
+++ b/.github/styles/Vocab/Codacy/accept.txt
@@ -15,6 +15,7 @@ Cobertura
 Codacy
 CodeNarc
 CoffeeLint
+config
 Cppcheck
 Credo
 CSSLint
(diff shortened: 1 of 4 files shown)


prcr:
@alerizzo This was the "simple solution" I talked about when we were discussing the feasibility of using LaunchDarkly on the docs.

@terciosoares For now, I propose using just one single feature flag that will enable us to hide beta features on Self-hosted releases. What is your opinion on this?

inesgomas (approved):

inesgomas:
It's weird to use the word "beta" to separate Cloud from SH, when we have BETA label for features released that "are being tested" it might happen to release something with the beta label to SH, like the standards. But it does the trick 🤷

prcr:
> It's weird to use the word "beta" to separate Cloud from SH, when we have BETA label for features released that "are being tested" it might happen to release something with the beta label to SH, like the standards. But it does the trick shrug

Yes, I also thought about that and I agree. But for now, I think it matches better the use case that we're trying to solve, so I don't mind updating the name (or creating more flags) if it's needed in the future. :pray:

PR #1087, opened 2022-03-02
toolchain: Clean up fixed Python packages for compatibility

prcr:


requirements.txt
```diff
@@ -12,8 +12,3 @@ mkdocs-include-markdown-plugin==3.3.0
 mkdocs-rss-plugin==0.21.0
 mkdocs-meta-descriptions-plugin==1.0.2
 mkdocs-exclude==1.0.2
-
-# The requirements below are fixed for compatibility
-Jinja2==2.11.3
```
prcr, line 17: Jinja2 is no longer incompatible with the latest version of mkdocs-rss-plugin (see https://github.com/codacy/docs/pull/769).

requirements.txt
```diff
@@ -12,8 +12,3 @@ mkdocs-include-markdown-plugin==3.3.0
...
-mkdocs-material<=7.2.3
```
prcr, line 18: We're now using mkdocs 1.2, which was a requirement for newer versions of mkdocs-material (see https://github.com/codacy/docs/pull/779).

requirements.txt
```diff
@@ -12,8 +12,3 @@ mkdocs-include-markdown-plugin==3.3.0
...
-MarkupSafe==2.0.1
```
prcr, line 19: By using the latest version of Jinja2 we can also use the latest version of MarkupSafe.

PR #1096, opened 2022-03-10
clean: Clarify "Last updated" column in Repositories list page CY-5784

prcr:
In the context of [CY-5784](https://codacy.atlassian.net/browse/CY-5784) we realized that the documentation didn't mention the exact value that the column **Last updated** on the Repository list page displays for each repository.

This pull request clarifies that the **Last updated** date is actually the date of the last (pushed) commit to the repository.

See also https://github.com/codacy/codacy-website/pull/125.

### :eyes: Live preview
https://deploy-preview-1096--docs-codacy.netlify.app/organizations/organization-overview/#last-updated-repositories

### :construction: To do
- [x] Wait until [CY-5784](https://codacy.atlassian.net/browse/CY-5784) is released


docs/organizations/managing-repositories.md
```diff
@@ -11,7 +11,7 @@ This page lists the repositories that already belong to your organization on Cod
 -   Complexity metric
 -   Duplication metric
 -   Coverage metric
--   Last updated time
+-   Date of the last commit
```
prcr, line 14: I'm guessing that this is actually the date of the last commit **in one of the enabled branches on Codacy**. Can anyone confirm this so we can also add this extra detail?

prcr: In lieu of the newest information, I now believe that the date is of the last commit pushed to **any** branch.

vascorsd (commented):
Would it be worth consider start talking about "Last Activity" instead of last updated and then trying to say that it's either about last commits / last pushes / last changes on the repo ?

prcr:
> Would it be worth consider start talking about "Last Activity" instead of last updated and then trying to say that it's either about last commits / last pushes / last changes on the repo ?

I do like this idea, as the UI copy would become a bit more precise. :+1:

On the other hand, we need to make sure that all places on the Codacy UI showing this information are updated, for example the Organization Overview:

image

The API would still mention `lastUpdated`, though (see https://github.com/codacy/codacy-website/pull/125). Which makes me consider that it could be best to leave the label as is in the scope of this fix. :thinking:

prcr:
Since the "Last updated" value depends on the implementation of each Git provider, I opted to include more details in the section [Last updated repositories](https://deploy-preview-1096--docs-codacy.netlify.app/organizations/organization-overview/#last-updated-repositories) of the Organization overview, and link the description of the columns of the repositories list to these details.

Please help confirm that the name of the fields and APIs mentioned are indeed the correct ones.

PR #1130, opened 2022-03-24
clean: Clarify that GitLab owner role is for groups CY-5882

prcr:
GitLab roles can be [for projects](https://docs.gitlab.com/ee/user/permissions.html#project-features-permissions) or [for groups](https://docs.gitlab.com/ee/user/permissions.html#group-members-permissions) (equivalent to GitHub's organizations).

This pull request clarifies that the "Owner" role mentioned is the **group-level** role.

### :eyes: Live preview
https://clean-clarify-gitlab-roles-cy-5882--docs-codacy.netlify.app/organizations/roles-and-permissions-for-synced-organizations/#permissions-for-gitlab

### :construction: To do
- [x] Merge https://github.com/codacy/docs/pull/1171 and rebase `master`

### DIFF
--- a/docs/organizations/roles-and-permissions-for-synced-organizations.md
+++ b/docs/organizations/roles-and-permissions-for-synced-organizations.md
@@ -151,7 +151,7 @@ The table below maps the GitLab Cloud and GitLab Enterprise roles to the Codacy 
       <td>No</td>
     </tr>
     <tr>
-      <th>Guest</th>
+      <th>Project Guest</th>
       <td class="yes">Yes<sup>2</sup></td>
       <td class="yes">Yes</td>
       <td><a href="#change-analysis-configuration">Configurable</a></td>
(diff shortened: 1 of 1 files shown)


prcr:
@lventura-codacy, do you think that you can confirm if my proposed changes are correct with little effort? If not, I can ask for help from the team.

prcr:
Feedback from @machadoit: we should also make explicit what is the behavior for **project Owner** (should be the same as project Maintainer).

I'll consider some alternatives to present a new proposal that will clarify these details while keeping (or improving) the readability of the table.

prcr:
@machadoit please have a look at the new version.

After thinking about this some more time, I reached the conclusion that it's best to be explicit and clearly mark if each role is for GitLab Projects or Groups, similar to what we were already doing for GitHub. For completeness and added clarity, I also added a new row for **Project Owner** with the same permissions as **Project Maintainer**.

image

machadoit:
SGTM! ;D

PR #1143, opened 2022-03-28
add supported includes on dartanalyzer tool, lints and flutter_lints CY-5626

rubensantoscodacy:


docs/getting-started/supported-languages-and-tools.md
```diff
@@ -380,7 +380,7 @@ The table below lists all languages and frameworks that Codacy supports and the
 
 <sup><span id="client-side">1</span></sup>: Supported as a [client-side tool](../related-tools/local-analysis/client-side-tools.md).  
 <sup><span id="cppcheck-misra">2</span></sup>: Currently, Cppcheck only supports [checking the MISRA guidelines for C](https://cppcheck.sourceforge.io/misra.php){: target="_blank"}.  
-<sup><span id="dart-limitations">3</span></sup>: Currently, Codacy doesn't support `include` statements in dartanalyzer configuration files, but you can [use a workaround to overcome this limitation](https://github.com/codacy/codacy-dartanalyzer#include-not-working-on-analysis_optionsyaml).  
+<sup><span id="dart-limitations">3</span></sup>: Currently, Codacy only supports following `include` on analysis_options.yaml file: [lints](https://pub.dev/packages/lints) and [flutter_lints](https://pub.dev/packages/flutter_lints).
```
prcr, line 383: Here's a way of making the text flow a bit more naturally. It also opens the external links on a new browser window.

```suggestion
<sup><span id="dart-limitations">3</span></sup>: Currently, Codacy only supports including the packages [lints](https://pub.dev/packages/lints){: target="_blank"} and [flutter_lints](https://pub.dev/packages/flutter_lints){: target="_blank"} on dartanalyzer configuration files.  
```

prcr: @rubensantoscodacy, just to confirm, the workaround that we had previously is no longer applicable, is this right?

prcr (changes requested):

prcr:
@rubensantoscodacy Is this change still related to the Jira issue [CY-5626](https://codacy.atlassian.net/browse/CY-5626)? Let's include the relevant Jira issue in the pull request title for future reference :+1:

docs/getting-started/supported-languages-and-tools.md
rubensantoscodacy: At least for supported lints/packages that we support include. I decided to remove the limitation since, there is no longer a limitation on our side, and if the client wants to use a lint/package that we do not support, either request the support (or do a PR on dartanalyzer 😛  ), or do the workaround.

rubensantoscodacy: What we can do, is to rename the previous link/section from limitation to **workaround on unsupported lints/pacakges**

prcr: I understand better now, the process should be similar to the one for ESLint plugins, for example. :+1:

So I suggest that leave it like this for now, but if we start supporting more lints packages we'll eventually need to remove this note as we should only maintain the list of supported packages directly on the [codacy/codacy-dartanalyzer](https://github.com/codacy/codacy-dartanalyzer/) repository.

prcr (approved):

PR #1170, opened 2022-04-06
toolchain: Use lowercase branch names for branch previews

prcr:
Apparently, using Netlify aliases with capital letters causes issues making the branch preview deployment unavailable. This ensures that the branch names uses as aliases are lowercase.


prcr:
See https://github.com/nwtgck/actions-netlify/issues/243 for a potential future fix that will allows us to simplify the GitHub Actions workflow.

PR #1181, opened 2022-04-13
toolchain: Run Vale on push instead of on pull_request

prcr:
This can possibly fix https://github.com/errata-ai/vale-action/issues/57.


prcr:
This change still doesn't fix the issue. :frowning_face:

PR #1179, opened 2022-04-08
feature: Map Codacy permission levels to Git provider roles DOCS-348

prcr:
Adds a column to the tables on the "Roles and permissions" page to list the Codacy permission levels that correspond to each role on the Git providers.

This extra mapping is necessary because users need to know the correspondence between the Codacy permission levels and the roles on their Git providers while [configuring the permission levels for changing the analysis configuration](https://feature-map-codacy-permission-level--docs-codacy.netlify.app/organizations/roles-and-permissions-for-synced-organizations/#change-analysis-configuration).

Fixes #1000.

### :eyes: Live preview
https://feature-map-codacy-permission-level--docs-codacy.netlify.app/organizations/roles-and-permissions-for-synced-organizations/

PR #1196, opened 2022-04-22
clean: Transpose roles and permissions tables DOCS-368

prcr:
The tables in the [Roles and permissions page](https://docs.codacy.com/organizations/roles-and-permissions-for-synced-organizations/) have become increasingly harder to read as we added more columns for Codacy operations and for the Codacy permission levels (see https://github.com/codacy/docs/pull/1179).

By transposing the tables:

- The columns contain each Git provider role

    The number of roles on each Git provider is not expected to increase, and the names of the roles are much shorter than the Codacy operations. This means that the tables don't need to be as wide.

- There is a row for each Codacy operation

    This makes it much easier to add new operations if necessary without worrying about the columns becoming too "squished" together.

Before:

image

After:

image

## :eyes: Live preview
https://clean-transpose-permissions-tables--docs-codacy.netlify.app/organizations/roles-and-permissions-for-synced-organizations/

## :construction: To do
- [x] Review all tables to check if they're correct
- [x] Merge https://github.com/codacy/docs/pull/1179 and rebase `master`

### DIFF
--- a/docs/organizations/roles-and-permissions-for-synced-organizations.md
+++ b/docs/organizations/roles-and-permissions-for-synced-organizations.md
@@ -97,7 +97,7 @@ The table below maps the GitHub Cloud and GitHub Enterprise roles to the corresp
       <td class="yes">Yes</td>
     </tr>
     <tr>
-      <td>Invite and accept members,<br/>modify billing"</td>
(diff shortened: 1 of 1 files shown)


machadoit:
Fair enough, as a costumer that want to consume this information, seems better 👌

PR #1206, opened 2022-04-27
toolchain: Enable Netlify deployment comments

prcr:
Enable pull request comments containing the Netlify branch preview deployment URL.


prcr:
I gave up on trying to set up the comments because the workflow would have to run for the `pull_request` event.

PR #1233, opened 2022-05-06
feature: Introduce concept of not applicable diff coverage CY-5998

prcr:
Updates all relevant pages to mention the new possible value `∅` (not applicable) for diff coverage.

### :eyes: Live preview
- https://feature-diff-coverage-not-applicabl--docs-codacy.netlify.app/repositories/pull-requests/#pull-request-quality-overview
- https://feature-diff-coverage-not-applicabl--docs-codacy.netlify.app/repositories-configure/adjusting-quality-settings/#gates
- https://feature-diff-coverage-not-applicabl--docs-codacy.netlify.app/faq/code-analysis/which-metrics-does-codacy-calculate/#code-coverage


docs/repositories-configure/adjusting-quality-settings.md
```diff
@@ -27,7 +27,7 @@ The following screenshot displays the default configuration values:
 -   **New security issues are over:** Pull requests or commits are marked not up to standards if the number of security issues introduced is higher than the set value.
 -   **Complexity is over:** Pull requests or commits are marked not up to standards if the introduced complexity is higher than the set value.
 -   **Duplication is over:** Pull requests or commits are marked not up to standards if the number of clones introduced is higher than the set value.
--   **Diff coverage is under:** Pull requests are marked not up to standards if the diff coverage of the pull request is lower than the set value. This rule is only available for pull requests.
+-   **Diff coverage is under:** Pull requests are marked not up to standards if the diff coverage of the pull request is applicable (different from `∅`) and lower than the set value. This rule is only available for pull requests.
```
fjrdomingues, line 30: I'm thinking that mentioning the Null here convolutes the explanation of what the gate does. Makes it harder to understand how the gate works. I would wonder what "applicable" means and what use-cases cause that.

prcr: Goo point, I was also unsure if I should mention it here or not. But I would also like to mention somewhere that the `∅` value doesn't make the quality gate fail. :thinking: Do you think this is relevant? Or that perhaps, since this is actually the expected thing to happen, it's not worth mentioning it for now?

docs/faq/code-analysis/which-metrics-does-codacy-calculate.md
```diff
@@ -112,7 +112,7 @@ Codacy calculates code coverage as follows:
 -   The coverage value for each file is the percentage of coverable lines that are covered by tests in the file
 -   A repository is considered to have acceptable coverage if the average coverage value for the files in the repository is higher than the threshold [**Coverage is under**](../../repositories-configure/adjusting-quality-settings.md#goals)
 -   The coverage variation value of a commit or pull request is the number of percentage points that the coverage value for the file increased or dropped in the commit or pull request
--   The diff coverage of a pull request is the percentage of coverable lines that were added or modified in the pull request that are covered by tests
+-   The diff coverage of a pull request is the percentage of coverable lines added or modified in the pull request that are covered by tests, or `∅` (not applicable) if no coverable lines were changed
```
machadoit, line 115: The sentence starts to be a bit long/difficult to me. Not sure if the `, or` could be a note or a sub-bullet? But maybe that is worse 😅 . Either way, it is understandable, just something that starts to be a bit too big for my *sentence comprehension buffer* :D

docs/repositories-configure/adjusting-quality-settings.md
machadoit: I tend to agree, but at the same time, it would be nice to have some kind of place with this info, like you just did. So for me, we can either leave it as it is, or consider linking the `applicable` to the place where we explain the `**Diff coverage:** Code coverage of the coverable lines added or changed by the pull request, or `∅` if no coverable lines were changed`

docs/repositories/commits.md
```diff
@@ -46,7 +46,7 @@ This area displays the quality gate status and an overview of the code quality m
 {% endif %}
 {% if page.meta.page_name == "pull request" %}
     -   **Coverage variation:** Variation of code coverage relative to the target branch
-    -   **Diff coverage:** Code coverage of the lines added or changed by the pull request
+    -   **Diff coverage:** Code coverage of the coverable lines added or changed by the pull request, or `∅` if no coverable lines were changed
```
machadoit, line 49: Could we make the `applicable` word show here somehow? What I am thinking is that I would love to have a reference place when people have either of 2 questions?
* What does it mean to be applicable?
* What does the ø means?

What do you think?

machadoit (approved):
Although I am giving some comments, I think for the current structure we can not make crystal clear everywhere it seems. I think unless we want to have a place dedicated to DiffCoverage and always want to link to there or something, it will be always hard, because it is a special case that we will keep feeling the need of mentioning.

That said, I like it and it makes the information available, so for me, it is good to go! :D

docs/repositories-configure/adjusting-quality-settings.md
prcr: I think that both your feedback is valid.

For now, I decided to simplify a bit the description of the diff coverage quality gate, while still mentioning that the value `∅` also passes the gate, and including a link on "not applicable" to the section where we define the code coverage metrics.

Tell me what you think, as we still have some time to reconsider this until we release the feature.

docs/repositories/commits.md
prcr: This makes a lot of sense, yes. We're now referring to the new diff coverage value more consistently as "`∅` (not applicable)". :+1:

fjrdomingues (approved):

PR #1274, opened 2022-06-07
feature: Update supported GitLab Self-managed version to 14.8

prcr:
Updates the supported GitLab Self-managed version to [14.8](https://docs.gitlab.com/ee/update/#1480).

### :eyes: Live preview
https://feature-update-gitlab-self-managed--docs-codacy.netlify.app/faq/general/which-version-control-systems-does-codacy-support/

### DIFF
--- a/docs/faq/general/which-version-control-systems-does-codacy-support.md
+++ b/docs/faq/general/which-version-control-systems-does-codacy-support.md
@@ -39,7 +39,7 @@ Codacy supports repositories from the following Git providers:
       <td><p>Codacy Cloud or<br/><a target="_blank" href="https://www.codacy.com/self-hosted">Codacy Self-hosted</a></p>
     </tr>
     <tr>
-      <td><p><strong>GitLab Self-managed</strong><br/>version 12.6.2-ee or later</p></td>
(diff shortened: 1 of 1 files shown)


docs/faq/general/which-version-control-systems-does-codacy-support.md
```diff
@@ -39,7 +39,7 @@ Codacy supports repositories from the following Git providers:
       <td><p>Codacy Cloud or<br/><a target="_blank" href="https://www.codacy.com/self-hosted">Codacy Self-hosted</a></p>
     </tr>
     <tr>
-      <td><p><strong>GitLab Self-managed</strong><br/>version 12.6.2-ee or later</p></td>
+      <td><p><strong>GitLab Self-managed</strong><br/>version 14.0.0 or later</p></td>
```
prcr, line 42: I checked the GitLab version on gitlab.dev.codacy.org and I saw it is actually 14.8. Should we use that version instead, or is 14.0.0 fine?

image

prcr: Makes sense @lolgab. I updated the version number accordingly. :+1:

PR #1373, opened 2022-08-25
feat: Mention persistence of Organization Overview filter CY-5651

prcr:
Mentions that the Organization Overview filter configuration is stored in-browser to persist the configuration between visits to the page.

In this pull request, I also decided to remove the "tip" card and just mention the languages filter in the main flow of the page. Like this, the screenshot illustrates better not only the languages filter but also the look and feel of the entire filter drop-down list.

### :eyes: Live preview
https://feat-org-overview-filter-persistenc--docs-codacy.netlify.app/organizations/organization-overview/


docs/organizations/organization-overview.md
```diff
@@ -23,10 +23,9 @@ To access your Organization Overview, select an organization from the top naviga
 
 Use the drop-down list at the top of the page to filter the information displayed on all dashboard areas based on the repositories that you select. For example, you can use the filter to monitor the quality of the repositories maintained by specific teams or that include certain programming languages, or to ignore legacy repositories that are no longer maintained.
 
-!!! tip
-    Use the language filter to help you narrow down the list of repositories in the drop-down list:
+You can use the language filter to help you narrow down the list of repositories in the drop-down list, and the selected repositories are stored in your browser so that the same filter is applied between your visits to the Organization Overview page.
```
prcr, line 26: @alerizzo When I read your message announcing the new feature I realized that the docs didn't mention the persistence of the filter configuration, which is a nice feature. What do you think of this tweak?

inesgomas: Having this here makes it seem like it is only regarding the language filter. When it is more about "we keep your repositories filter". What about mention the persistence in the paragraph above? 👀

prcr: Good point @inesgomas! I made the change you suggest. :ok_hand:

inesgomas (approved):

PR #1476, opened 2022-10-20
feat: Add Unity as a supported framework by codacy-roslyn IO-100

prcr:
Updates the documentation with support for the Unity framework using [codacy/codacy-roslyn](https://github.com/codacy/codacy-roslyn).

### :eyes: Live preview
- https://feat-add-roslyn-tool-io-100--docs-codacy.netlify.app/getting-started/supported-languages-and-tools/
- https://feat-add-roslyn-tool-io-100--docs-codacy.netlify.app/repositories-configure/configuring-code-patterns/#using-your-own-tool-configuration-files
- https://feat-add-roslyn-tool-io-100--docs-codacy.netlify.app/repositories-configure/codacy-configuration-file/
- https://feat-add-roslyn-tool-io-100--docs-codacy.netlify.app/related-tools/local-analysis/client-side-tools/
- https://feat-add-roslyn-tool-io-100--docs-codacy.netlify.app/related-tools/codacy-plugin-tools/

### :construction: To do
- [x] Validate the correct name and URL for the new tool
- [x] Review and update the [tool description](https://github.com/codacy/codacy-roslyn/blob/master/docs/tool-description.md)
- [x] Check if the new tool reports security issues and update `docs/repositories/security-monitor.md`
    - There aren't any security code patterns in https://github.com/codacy/codacy-roslyn/blob/master/docs/patterns.json
- [x] Solve all `TODO` comments


docs/getting-started/supported-languages-and-tools.md
```diff
@@ -340,6 +340,13 @@ The table below lists all languages and frameworks that Codacy supports and the
       <td><a href="https://github.com/kucherenko/jscpd">jscpd</a></td>
       <td><a href="https://eslint.org/">ESLint</a></td>
     </tr>
+    <tr>
+      <td>Unity</td>
+      <td><a href="https://github.com/dotnet/roslyn-analyzers">Roslyn Analyzers</a><a href="#client-side"><sup>1</sup></a></td>
```
prcr, line 345: Do you agree that the new tool name is "Roslyn Analyzers" and that the tool URL is https://github.com/dotnet/roslyn-analyzers?

docs/related-tools/local-analysis/client-side-tools.md
```diff
@@ -88,6 +88,12 @@ The table below describes the supported client-side tools and includes links to
         <td>Faux Pas inspects your iOS or Mac app's Xcode project and warns about possible bugs, as well as about maintainability and style issues.</td>
         <td><a href="https://github.com/codacy/codacy-faux-pas#usage">Running Faux Pas</a> (standalone)</td>
     </tr>
+    <tr>
+        <td>Unity</td>
+        <td><a href="https://github.com/dotnet/roslyn-analyzers">Roslyn Analyzers</a></td>
+        <td><!--TODO Add final description--></td>
```
prcr, line 94: We're missing the tool description (https://github.com/codacy/codacy-roslyn/blob/master/docs/tool-description.md). Please make a first proposal and I'll review it.

docs/getting-started/supported-languages-and-tools.md
pedrocodacy: yup

docs/getting-started/supported-languages-and-tools.md
```diff
@@ -340,6 +340,13 @@ The table below lists all languages and frameworks that Codacy supports and the
       <td><a href="https://github.com/kucherenko/jscpd">jscpd</a></td>
       <td><a href="https://eslint.org/">ESLint</a></td>
     </tr>
+    <tr>
+      <td>Unity</td>
```
vascorsd, line 344: We should be saying that the language is C# and Unity is one of the "plugins" we are supporting currently.

docs/related-tools/local-analysis/client-side-tools.md
```diff
@@ -88,6 +88,12 @@ The table below describes the supported client-side tools and includes links to
         <td>Faux Pas inspects your iOS or Mac app's Xcode project and warns about possible bugs, as well as about maintainability and style issues.</td>
         <td><a href="https://github.com/codacy/codacy-faux-pas#usage">Running Faux Pas</a> (standalone)</td>
     </tr>
+    <tr>
+        <td>Unity</td>
```
vascorsd, line 92: Same point as above about the tool being for the C# language.

vascorsd (changes requested):

docs/getting-started/supported-languages-and-tools.md
prcr: My idea was to associate the Rosnlyn Analyzers with the Unity framework (notice that on the [Supported languages and tools](https://feat-add-roslyn-tool-io-100--docs-codacy.netlify.app/getting-started/supported-languages-and-tools/) the first column is called "Language or framework" and that we have entries there like Kubernetes or Terraform). This would give more visibility to the fact that we support analyzing Unity code (in C#).

However, I made a new proposal - that I'm still not fully convinced about. :sweat_smile: What do you think?

vascorsd: That doesn't seem wrong either, since the tool right now **only** supports unity rules, but the tool is generic enough to later be changed to include other rules. Like eslint that supports multiple plugins to support more stuff. 

I would maybe fall on the side of keeping it in the C# language since the tool will is associated to the C# language files so a user will see the tool recommended in the UI in the Code Patterns page when they simply have c# files... But I'm totally open to another approach here.

prcr: Perhaps @fjrdomingues has an opinion on this.

vascorsd (approved):
:100:

PR #1500, opened 2022-11-08
feat: Update behavior when using a coding standard PLUTO-98

claudiacarpinteiro:
Updates the product behavior when a repository follows a coding standard:

-   The repository now links to the coding standard configurations, instead of copying/overwriting the configurations to the repository
-   Changes made to any tool or code pattern cause the repository to stop following the coding standard, and Codacy copies the coding standard configurations to the repository so users can customize them.

### :eyes: Live preview
- https://feat-update-following-coding-standa--docs-codacy.netlify.app/organizations/using-a-coding-standard/
- https://feat-update-following-coding-standa--docs-codacy.netlify.app/repositories-configure/configuring-code-patterns/
- https://feat-update-following-coding-standa--docs-codacy.netlify.app/repositories/issues/#ignoring-and-managing-issues

### :construction: To do
- [x] Wait until the feature gets to production

### DIFF
--- a/docs/organizations/using-a-coding-standard.md
+++ b/docs/organizations/using-a-coding-standard.md
@@ -46,8 +46,6 @@ To create a coding standard for your organization:
 
 1.  Select existing repositories that should follow the new coding standard and click **Save and apply standard**.
 
-    {% include "../assets/includes/coding-standard-apply-important.md" %}
-
     Codacy will start using the new coding standard on the next analysis of each selected repository.
 
     ![Applying the coding standard to repositories](images/coding-standard-apply.png)
@@ -85,8 +83,6 @@ To edit an existing coding standard or change the repositories that follow that 
     ![Renaming a coding standard](images/coding-standard-rename.png)
 
 1.  Click the button **Save and apply standard** on the repository selection page to save your changes to the coding standard.
-
-    {% include "../assets/includes/coding-standard-apply-important.md" %}
 
     Codacy will start using the updated coding standard on the next analysis of each selected repository.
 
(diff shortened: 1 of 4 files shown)


docs/organizations/using-a-coding-standard.md
```diff
@@ -86,8 +84,6 @@ To edit an existing coding standard or change the repositories that follow that
 
 1.  Click the button **Save and apply standard** on the repository selection page to save your changes to the coding standard.
 
-    {% include "../assets/includes/coding-standard-apply-important.md" %}
```
prcr, line 89: We should also delete the included file `docs/assets/includes/coding-standard-apply-important.md` since it seems like after these changes it will no longer be used. :broom:

claudiacarpinteiro: Yes, I was thinking of doing that, but I wanted to validate with you.

docs/organizations/using-a-coding-standard.md
```diff
@@ -86,6 +86,8 @@ To edit an existing coding standard or change the repositories that follow that
 
     Codacy will start using the updated coding standard on the next analysis of each selected repository.
 
+If you stopped applying the coding standard to any repository, Codacy will restore the previous code pattern configurations for that repository.
```
prcr, line 89: I wonder if we should give more visibility to this part using an `important` admonition (like we had for the previous warning that we're now removing).

Also, consider (with the team) if we should make it more explicit that Codacy restores the code pattern configurations **only** but not the tool status. :thinking:

prcr (approved):

PR #1502, opened 2022-11-11
fix: Integrate submodules to fix anchor link offset - IO-233

nicklem:
Integrates updates on submodules


prcr:
This was superseded by #1503.

PR #1532, opened 2022-12-01
fix: Vale - ignored block checked incorrectly

nicklem:
@prcr 

I'm creating this to fix an issue we noticed with Vale incorrectly detecting ignored blocks:

```
<!-- vale off -->

text here would sometimes not be ignored!

<!-- vale on -->
```

Upon opening a pull request on the Vale repo, the author suggested that I send over the `.vale.ini` file, where I noticed a suspicious regex match under `TokenIgnores`, `<.*>`, which would match HTML comments too.

Here's an attempted fix. It does fix our ignore issue, but I'm not quite sure what that regex matching rule was there for initially, so some context could be useful for a final sanity check before merging this one.

Thanks!

### DIFF
--- a/.vale.ini
+++ b/.vale.ini
@@ -7,7 +7,7 @@ Vocab = Codacy
 
 [*.md]
 # Ignore Markdown includes and bare URLs
-TokenIgnores = ({%\s*include.*%})|(<.*>)|({:.*})
+TokenIgnores = ({%\s*include.*%})|(<(?!\!--).*>)|({:.*})
 BlockIgnores = (?s){%\s*include.*%}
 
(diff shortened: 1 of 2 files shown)


prcr:
I added the regex in #1376 `<.*>` while trying to ignore some stubborn Vale false positives inside HTML tags.

But I completely forgot that the rule would also catch HTML comments :facepalm:, and I remember that at the time I lost quite some time trying to figure out why the change that I now pushed on 9ab013958fde3a3ea6379dc3e98553832bca1a16 wouldn't work.

I'm glad you were finally able to figure this out! :raised_hands:

prcr (approved):
Good catch! :heart:

PR #1539, opened 2022-12-06
fix: Mobile header select max width

nicklem:
Fixes excessive header select width, which pushed the CTA off-screen.
See: 
* https://github.com/codacy/codacy-mkdocs-material/pull/71
* https://github.com/codacy/codacy-mkdocs-material/pull/69


prcr (approved):
Perfect, it's working now! :raised_hands: 

It was a simple solution after all, I'm sorry I didn't remember earlier that we could edit the styles of the version selector directly here.

image

PR #1545, opened 2022-12-14
feat: Add guidelines for encoding description lists

prcr:
Markdown doesn't support [description lists](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl), but we should have a convention for encoding information that fits the format of "term/details" pairs.

### :eyes: Live preview
https://github.com/codacy/docs/blob/feat/contributing-description-lists/CONTRIBUTING.md#markdown-conventions

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -75,6 +75,15 @@ Follow these Markdown conventions when editing the documentation:
         1. Item 2.1
     ```
 
+-   Use the following structure to encode information that would fit an [HTML description list](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl):
+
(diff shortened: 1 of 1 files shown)


nicklem (approved):
Pity we don't have description list support, but it's not used often...we can live with this convention for now. LGTM

prcr:
> Pity we don't have description list support, but it's not used often...we can live with this convention for now.

We do have quite a few occurrences in this repository at least (you can check by searching for the regex `-   \*\*.*:\*\*`). I do believe that the convention is "good enough" and a proper Markdown syntax wouldn't add too much value anyway.

claudiacarpinteiro (approved):

PR #1546, opened 2022-12-14
feat: Add coverage summaries to GitHub integration COV-16 COV-19

prcr:
### :construction: To do
-   [ ] If relevant, include the Jira issue key at the end of the pull request title
-   [ ] Perform a self-review of the changes
-   [ ] Fix any issues reported by the CI/CD


docs/repositories-configure/integrations/github-integration.md
```diff
@@ -42,17 +42,29 @@ Adds a report to your pull requests showing whether your pull requests and cover
...
+### Issue summaries
 
 Shows an overall view of the changes in the pull request, including new issues and metrics such as complexity and duplication. To enable this option, you must enable **Status checks** first.
 
-![Pull request summary on GitHub](images/github-integration-pr-summary.png)
+![Issue summary on GitHub](images/github-integration-pr-summary.png)
+
+### Coverage summaries
+
+Adds a pull request comment showing an overall view of the coverage metrics for the pull request, including details about the data that Codacy used to calculate the coverage variation and diff coverage metrics.
```
nicklem, line 59: This is more a question than feedback, as I'm quite unsure: would it make sense to use the present tense and say `...data that Codacy uses to calculate...` here?

docs/coverage-reporter/index.md
```diff
@@ -443,3 +447,15 @@ Follow these instructions to validate that your coverage setup is working correc
     -   Branch name and commit SHA-1 hash corresponding to the CI/CD output
     -   Test coverage report that you're uploading to Codacy
     -   Any other relevant information or screenshots of your setup
+
+## 4. Enabling the Git provider integration {: id="enabling-git-provider-integration"}
+
+After validating that the coverage setup is complete, Codacy recommends that you enable the integration with your Git provider to obtain the coverage status and summaries directly on pull requests.
```
nicklem, line 453:
```suggestion
After validating that the coverage setup is complete, enable the integration with your Git provider to obtain the coverage status and summaries directly on pull requests.
```
I'm unsure what we mean by "Codacy recommends that" here. Is there a prompt, or is this "our" suggestion? Would it make sense to remove this, so the sentence is more imperative?

docs/repositories-configure/integrations/github-integration.md
prcr: The coverage summary includes information about the number of coverable and covered lines that Codacy received in the report that was used to calculate the final metrics. I believe that using the past tense makes more sense here because it helps to clarify the process that Codacy uses to calculate the final metrics: the final metrics are a result of the data that the user uploaded.

docs/coverage-reporter/index.md
prcr: Here the idea is to encourage the users to activate the Git provider integration features, as they'll obtain more value from Codacy. However, that's not mandatory.

We could write instead something like:

> After validating that the coverage setup is complete, _optionally_ enable the integration [...]

However, this doesn't seem to convey the same urgency or "advantage to the users" as an actual recommendation. :thinking: What are your thoughts?

nicklem: OK, I get it now, and I agree it should be stronger than an `optionally`.

I wonder if we may use `we recommend that`? Or is that too "personal"?

prcr: The Microsoft Style Guide [recommends against that](https://learn.microsoft.com/en-us/style-guide/grammar/person#avoid-plural-first-person-we-us). But now that I read that section more carefully, I realize that the alternative "Codacy recommends [...]" is also not what the style guide recommends. :sweat_smile: 

Also, we already have lots of existing occurrences of "we recommend" in the docs, and only three "Codacy recommends".

So, for now I'm going to change the sentence to "We recommend" for the sake of consistency, and we'll aim to review this style consistently across the whole docs in a future opportunity. Would you like to create a Jira issue in our backlog for this kind of future review @nicklem?

nicklem: 👍 https://codacy.atlassian.net/browse/DOCS-497

docs/coverage-reporter/index.md
```diff
@@ -443,3 +447,15 @@ Follow these instructions to validate that your coverage setup is working correc
     -   Branch name and commit SHA-1 hash corresponding to the CI/CD output
     -   Test coverage report that you're uploading to Codacy
     -   Any other relevant information or screenshots of your setup
+
+## 4. Enabling the Git provider integration {: id="enabling-git-provider-integration"}
+
+After validating that the coverage setup is complete, we recommend that you enable the integration with your Git provider to obtain the coverage status and summaries directly on pull requests.
```
fjrdomingues, line 453: If this feature will be enabled by default, does this recommendation make sense?

docs/repositories-configure/integrations/github-integration.md
```diff
@@ -42,17 +42,31 @@ Adds a report to your pull requests showing whether your pull requests and cover
...
+
+When there are new coverage results Codacy updates the last coverage summary comment if it's included in the last 5 comments of the pull request, otherwise Codacy creates a new comment.
```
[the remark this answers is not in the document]
prcr: Indeed, this is a rather long sentence. I'm going to accept your suggestions and also split the sentence in two:

```suggestion
When there are new coverage results, Codacy updates the last coverage summary comment if it's included in the last 5 comments of the pull request. Otherwise, Codacy creates a new comment.
```

docs/coverage-reporter/index.md
prcr: Both the GitHub status checks and the new coverage summary [will be on by default on new repositories](https://codacy.atlassian.net/browse/COV-22). :+1: But I was thinking about highlighting the feature for customers with existing repositories, or even for customers who may have turned off the feature by chance before they set up the coverage.

Indeed, the ROI of having this recommendation here should be low and it's probably better to keep the coverage setup page a little bit simpler. What are your thoughts @fjrdomingues?

fjrdomingues: Following the flow of the documentation, and with "After validating that the coverage setup is complete". It sounded to me like that's the use case where a user is adding Coverage for the first time. Recommending the user enable the feature may be wasteful because, most likely, the feature will already be enabled. Either way, I think this is minor.

I'm more concerned with users knowing that they need to have quality settings enabled for coverage to receive summaries and status checks on the provider. Do you think that's coverage in the docs?

prcr: > Recommending the user enable the feature may be wasteful because, most likely, the feature will already be enabled.

Good point, and I agree. I reverted the changes to the page [Adding coverage to your repository](https://feat-coverage-summaries-cov-16-cov--docs-codacy.netlify.app/coverage-reporter/). :heavy_check_mark: 

> I'm more concerned with users knowing that they need to have quality settings enabled for coverage to receive summaries and status checks on the provider. Do you think that's coverage in the docs?

Yes, for the status checks the users must set up coverage **and** configure quality gates for coverage, so [we already had a note here](https://docs.codacy.com/repositories-configure/integrations/github-integration/#status-checks). For the coverage summaries, users only need to set up coverage (and if they also have a quality gate configured they'll get the OK/NOK emojis). For that, [we're adding this note](https://github.com/codacy/docs/pull/1546/files#diff-7f4e89a3c87e75816c07be0a47a92090b7833df324af55e6fe259ed6e9b85e9eR63-R64). :heavy_check_mark: 

I'm going to mark this conversation as solved for now, but we'll have another chance to review everything end-to-end, already including the final screenshots, before we publish the changes.

docs/repositories-configure/integrations/github-integration.md
```diff
@@ -42,17 +42,31 @@ Adds a report to your pull requests showing whether your pull requests and cover
...
+### Coverage summaries
+
+Adds a pull request comment showing an overall view of the coverage metrics for the pull request, including details about the data that Codacy used to calculate the coverage variation and diff coverage metrics.
+
+When there are new coverage results, Codacy updates the last coverage summary comment if it's included in the last 5 comments of the pull request. Otherwise, Codacy creates a new comment.
+
+!!! important
+    **To get coverage summaries** you must also [add coverage to your repository](../../coverage-reporter/index.md).
+
+![Coverage summary on GitHub](images/github-integration-coverage-summary.png)
```
LazyBun, line 66: Do you think that it would be worth it to show all possible versions of this notification in docs, or would that be too much?

LazyBun (approved):
LGTM

docs/repositories-configure/integrations/github-integration.md
prcr: That would definitely be too much - it would take too much space on the page in exchange for very little value.

I tend to prefer a more minimalist approach to docs, which works well when the UX is good and users can pretty much figure things out on their own. We can always add more details later if we get feedback suggesting that something's not clear or missing. :+1:

PR #1555, opened 2022-12-21
clean: Rename included Markdown block

prcr:
This is a small cleanup to make the name of an included Markdown block more explicit and parallel with the other existing block `<!--coverage-status-start--> [...] <!--coverage-status-end-->`.


nicklem (approved):
:shipit:

PR #1567, opened 2023-01-05
toolchain: Ignore Codacy analysis on Vale styles DOCS-479

prcr:
Excludes the path `.github/styles` from the Codacy analysis to prevent false positive issues on "3rd party" files.

### DIFF
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -1,2 +1,3 @@
 ---
-exclude_paths: []
+exclude_paths:
+  - ".github/styles/**"


nicklem (approved):

.codacy.yml
```diff
@@ -1,2 +1,3 @@
 ---
-exclude_paths: []
+exclude_paths:
+  - ".github/styles/**"
```
claudiacarpinteiro, line 3: @prcr I should do the same for Pulse, right?

prcr: Yes, that's where I got the idea to ignore the Vale styles since these are files that we typically don't edit directly.

I'm still testing if this works, but if it does, you can apply the same change on https://github.com/codacy/pulse-user-docs/pull/167. I'll let you know! :raised_hands:

prcr: Actually, I think it's better to test if this works on your pull request @claudiacarpinteiro. If after applying this change the issues reported by Codacy disappear, then it's working as expected.

PR #1577, opened 2023-01-11
feat: Multiple Coding Standards IO-352

nicklem:
Updates the page documenting the creation of a coding standard to account for the creation and management of multiple coding standards.

### :construction: To do
- [ ] Update instructions after the design is finalized (it's still evolving as I'm opening this)
- [ ] Screenshots
- [ ] ... TBD


prcr:
Hi @nicklem, I noticed that you're making a few improvements/edits that are unrelated to the multiple coding standards feature. :muscle: But I suggest that you perform those initial edits on a standalone pull request that Cláudia and I can quickly help you merge, and then rebase this pull request and continue working just on the changes directly related to multiple coding standards.

This will make life easier for the stakeholders that you later select as reviewers of the pull request because they won't be confused with the additional changes that are unrelated to the new feature.

nicklem:
Thanks for the tip @prcr! On it

claudiacarpinteiro (approved):
Looks good!
I believe you still have to replace the screenshots with the final ones, but I can approve already.

nicklem:
Thanks @claudiacarpinteiro!
Correct, the screenshots are done, but I'm still flagging them as temporary, pending any last-minute changes. Save for the red borders, as far as I'm concerned, these are final.

I'll await more feedback from the squad.

docs/organizations/using-a-coding-standard.md
```diff
@@ -5,24 +5,27 @@
...
 
-Create a coding standard on your organization to define and apply shared tool and code pattern configurations consistently across your repositories. You can also apply the coding standard to new repositories automatically by defining the coding standard as default.
+Create coding standards on your organization to define and apply shared tool and code pattern configurations consistently across your repositories. You can also apply a coding standard to new repositories automatically by defining it as default.
 
-The organization coding standard helps you ensure that Codacy analyzes multiple repositories with the same tool and code pattern settings. For example, you can use the organization coding standard to ensure that all your repositories and teams are following the same security rules or coding conventions.
+Coding standards help you ensure that Codacy analyzes multiple repositories with the same tool and code pattern configurations. For example, you can use a coding standard to ensure that a team or a group of repositories follow the same security rules or coding conventions.
```
albertoadami, line 10: I think that mention `team` here make the feature confusing, to me as a reader it seems like there is the possibility to associate a coding standard directly with a team, feature that we don't have at the moment.

Maybe will be nice to have something like this:
`Coding standards help you ensure that Codacy analyzes multiple repositories with the same tool and code pattern configurations. For example, you can use a coding standard to ensure that a group of repositories follow the same security rules or coding conventions.`

prcr (approved):
Apart from the feedback from @albertoadami, the updates look perfect to me - I'm already giving this an approve. Great work, @nicklem! :star_struck:

inesgomas (approved):

docs/organizations/using-a-coding-standard.md
nicklem: Sounds good, updating now.

```suggestion
Coding standards help you ensure that Codacy analyzes multiple repositories with the same tool and code pattern configurations. For example, you can use a coding standard to ensure that a group of repositories follow the same security rules or coding conventions.
```

docs/organizations/using-a-coding-standard.md
```diff
@@ -5,24 +5,27 @@
```
EmaTiba, line 5: In the paragraph below we still link o the old release announcements and the video showing the old version

docs/organizations/using-a-coding-standard.md
EmaTiba: I would be more specific in saying that the multiple repositories that Codacy analyses are the ones you selected:

`Coding standards help you ensure that Codacy analyzes the multiple repositories you selected / associated to a coding standard with the...`

docs/organizations/using-a-coding-standard.md
```diff
@@ -5,24 +5,27 @@
...
 
-The organization coding standard helps you ensure that Codacy analyzes multiple repositories with the same tool and code pattern settings. For example, you can use the organization coding standard to ensure that all your repositories and teams are following the same security rules or coding conventions.
+Coding standards help you ensure that Codacy analyzes multiple repositories with the same tool and code pattern configurations. For example, you can use a coding standard to ensure that a team or a group of repositories follow the same security rules or coding conventions.
 
-When you customize the tools or code patterns of a repository that follows the coding standard, Codacy warns you that the repository will stop following the coding standard and asks for your confirmation.
+When you customize the tools or code patterns of a repository that follows a coding standard, Codacy warns you that the repository will stop following the coding standard and asks for your confirmation.
```
EmaTiba, line 13: Maybe it's already understandable, but how about having something like this:

`When you customize the tools or code patterns, at the repository level, of a repository that`

EmaTiba: Do we need to link to the specific documentation?

docs/organizations/using-a-coding-standard.md
```diff
@@ -60,7 +63,10 @@ To set a coding standard as default:
 
 1.  Open your organization **Coding standards** page, tab **Coding standards**.
 
-1.  Toggle **Make default** on the coding standard card.
+1.  Toggle **Make default** on the relevant coding standard card.
+
+    !!! note
+        Only one coding standard at a time can be the default coding standard.
```
EmaTiba, line 70: Do we want to inform users that if they accidentally turn off the make default there's no way to automatically revert to the previous status?

docs/organizations/using-a-coding-standard.md
prcr: In my opinion this adds some extra complexity to the sentence with no extra clarity about how the feature works, as it's already implicit that only some repositories (the ones you select) are affected by each coding standard.

docs/organizations/using-a-coding-standard.md
prcr: As a rule of thumb, we should use links frugally because each link is like a potential "fork" in the flow of the user reading the docs. The goal of this sentence is to explain that the repositories become "unliked" from the coding standard if there's a change to the tool and code pattern configurations of the repository. Adding a link to the other page here can actually detract from this goal.

docs/organizations/using-a-coding-standard.md
prcr: If we go to this level of detail here, we should probably do the same for every other feature on the Codacy app that doesn't have a corresponding "undo" action - and that's practically all of them. :slightly_frowning_face: I wouldn't call the attention to this detail here.

prcr:
@EmaTiba My replies to your feedback may seem a bit "dismissive", but please don't interpret them that way. It's a hard thing to balance clarity on the docs against including a description of every single detail and edge case. :pray: 

It's best to start with a more minimalist approach and include only the most relevant information, that's easier to consume and understand by the users. After the docs are live, we constantly monitor the feedback that we receive on them and can then adjust if we learn that something isn't clear or the users actually need more info to perform their tasks.

docs/organizations/using-a-coding-standard.md
EmaTiba: Seeing it this way it makes sense to keep the focus on the behaviour and not have a link.

docs/organizations/using-a-coding-standard.md
prcr: I've now reviewed both the blog post announcement and the video more carefully:

- The video is generic enough and the features described by Harry still exist, so I think it's still worthwhile keeping it
- The blog post on the other hand mentions some limitations that we're actually overcoming with this update. Maybe there will be a new announcement that we can link to from the docs? :rocket:

nicklem: There's a [follow-up task for](https://codacy.atlassian.net/browse/IO-317) this. I'll link this thread there 😉

docs/organizations/using-a-coding-standard.md
nicklem: I second Paulo's suggestion. Resolving this conversation

PR #1580, opened 2023-01-12
toolchain: Fix generating robots.txt file DOCS-455

prcr:
The changes from the previous pull request https://github.com/codacy/docs/pull/1578 didn't work because we were creating the `robots.txt` file in the wrong directory. The idea is to create the file inside the directory `./site` where all the generated HTML files are so that it's pushed together with them to the branch `gh-pages`.

Besides that, I had copied Pulse's domain name by mistake! :sweat_smile:

### DIFF
--- a/.github/workflows/mkdocs.yml
+++ b/.github/workflows/mkdocs.yml
@@ -74,12 +74,12 @@ jobs:
         run: |
           echo "MKDOCS_VERSION=$(mkdocs --version | cut -d " " -f 3)" >> $GITHUB_ENV
           echo "SHORT_SHA=`echo ${GITHUB_SHA} | cut -c1-9`" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


nicklem (approved):
I should pay more attention when checking code edits...

claudiacarpinteiro (approved):
I'll wait for this merge before I merge [#173](https://github.com/codacy/pulse-user-docs/pull/173) 😉

prcr:
> I should pay more attention when checking code edits...

What counts is that we can fail fast! :rocket:

By the way, the fact that the domain name was wrong was what allowed me to understand that the previous version wasn't working because the domain didn't change in [the commit created by running the workflow](https://github.com/codacy/docs/commit/3053915fa2a8fb66e48efe276a9fcf5afe8f2df2). :laughing:

PR #1596, opened 2023-01-25
feat: API example to check for PR commits with missing coverage data DOCS-503

prcr:
Adds an example on how to use the Codacy API endpoint [getPullRequestCoverageReports](https://api.codacy.com/api/api-docs#getpullrequestcoveragereports) to determine if Codacy is missing coverage data that is required to calculate the coverage metrics for pull requests.

### :eyes: Live preview
https://feat-api-example-pr-reports-docs-50--docs-codacy.netlify.app/codacy-api/examples/identifying-commits-without-coverage-data/

### DIFF
--- a/docs/codacy-api/examples/identifying-commits-without-coverage-data.md
+++ b/docs/codacy-api/examples/identifying-commits-without-coverage-data.md
@@ -49,7 +49,10 @@ while read pull_request_number; do
 done
 ```
 
-Example usage and output:
+Example usage and output, where:
+
+-   The first commit listed for each pull request is the **head commit** of the pull request branch
+-   The second commit listed for each pull request is the **common ancestor commit** of the pull request branch
 
 ```bash
 $ ./check-coverage.sh pulse
--- a/docs/codacy-api/examples/identifying-commits-without-coverage-data.md
+++ b/docs/codacy-api/examples/identifying-commits-without-coverage-data.md
@@ -48,7 +48,7 @@ done
 Example output:
 
 ```bash
-$ ./run.sh pulse
+$ ./check-coverage.sh pulse
 Checking #1563
 Coverage for 4faccc86676f7dba3af2b71400605b0be4a686e3 is Processed
 Coverage for 51e57784468459b9b2839aa63c3e7e807a39c4ab is null
--- a/docs/codacy-api/examples/identifying-commits-without-coverage-data.md
+++ b/docs/codacy-api/examples/identifying-commits-without-coverage-data.md
@@ -19,7 +19,16 @@ To check if Codacy has received the required coverage data to calculate the cove
 
 ## Example: Identifying which pull request commits are missing coverage data
 
-<!--TODO Explain example-->
+This example checks whether the head and common ancestor commits of the open pull requests currently in a repository have received coverage data.
+
+The example script:
+
+
+1.  Defines the [account API token](../api-tokens.md#account-api-tokens) used to authenticate on the Codacy API, the Git provider, the organization name, and the repository name passed as an argument to the script.
+1.  Calls the Codacy API endpoint [listRepositoryPullRequests](https://api.codacy.com/api/api-docs#listrepositorypullrequests) to retrieve the list of open pull requests on the repository.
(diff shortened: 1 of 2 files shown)


docs/codacy-api/examples/identifying-commits-without-coverage-data.md
```diff
@@ -0,0 +1,73 @@
...
+To calculate the supported coverage metrics for pull requests, Codacy requires that at least the following commits provide coverage data:
+
+-   The common ancestor commit of the pull request branch and the target branch
+-   The head commit of the pull request branch
+
+The following diagram highlights the commits that must have received coverage data for Codacy to display the coverage variation metric on a pull request:
+
+![Commits that must have received coverage data](../../coverage-reporter/images/coverage-pr-commits.png)
+
+However, different factors may prevent your setup from correctly reporting coverage data for the required commits.
```
nicklem, line 16: If I understand correctly, these different factors are all in the hands of the customer. That is, **they** must set everything up correctly.

The way it's phrased, a user may read "different factor" as potentially something internal to Codacy (at least I did). I'd rephrase it to clarify that Codacy is at the receiving end and expected to work. Something like:

```suggestion
However, a misconfigured setup may fail to report coverage data to Codacy for the required commits.
```

nicklem (approved):
Left a minor note. Otherwise, LGTM! :shipit:

docs/codacy-api/examples/identifying-commits-without-coverage-data.md
```diff
@@ -0,0 +1,73 @@
...
+
+1.  Defines the [account API token](../api-tokens.md#account-api-tokens) used to authenticate on the Codacy API, the Git provider, the organization name, and the repository name passed as an argument to the script.
+1.  Calls the Codacy API endpoint [listRepositoryPullRequests](https://api.codacy.com/api/api-docs#listrepositorypullrequests) to retrieve the list of open pull requests on the repository.
+1.  Uses [jq](https://github.com/stedolan/jq) to select only the numbers that identify the pull requests on the Git provider.
+1.  For each pull request, outputs the pull request number and calls the Codacy API endpoint [getPullRequestCoverageReports](https://api.codacy.com/api/api-docs#getpullrequestcoveragereports) to obtain the information about the coverage data received for the head and common ancestor commits of the pull request.
+1.  Uses [jq](https://github.com/stedolan/jq) to select and output the commit SHA-1 and coverage status for the commits.
+
+```bash
+CODACY_API_TOKEN="<your account API token>"
+GIT_PROVIDER="<your Git provider>" # gh, ghe, gl, gle, bb, or bbe
```
LazyBun, line 34: Would it make sense to explain the abbreviations?

pedrobpereira (approved):

docs/codacy-api/examples/identifying-commits-without-coverage-data.md
prcr: If you [check the troubleshooting table](https://docs.codacy.com/coverage-reporter/#status-commit-not-found) for the possible error status (which is not exhaustive) you'll probably agree that not every scenario is under the user's control. So I feel more comfortable with leaving the more ambiguous wording here.

docs/codacy-api/examples/identifying-commits-without-coverage-data.md
prcr: We do have a couple of API example pages that [follow a format that "handholds" the users a bit more](https://docs.codacy.com/codacy-api/examples/adding-people-to-codacy-programmatically/#adding-people-to-codacy-programmatically). However, both on the other examples and the [API reference](https://api.codacy.com/api/api-docs#getpullrequestcoveragereports) docs we don't have the explicit mapping but also never had reports of users not figuring this out. :thinking: 

For now, I'd prefer to keep the page a bit simpler but your feedback makes a lot of sense and if we ever understand that users have trouble understanding this we'll add the extra info across all examples for consistency. :+1:

prcr:
I'm going to publish the current version, but we'll have the opportunity to keep on improving the example if we have more insights.

PR #1597, opened 2023-01-25
clean: Simplify references to API endpoints in examples DOCS-503

prcr:
These are very minor edits to improve the consistency and simplify referring to API endpoints in the examples, which I noticed while starting to work on https://github.com/codacy/docs/pull/1596.


nicklem (approved):
LGTM. Would Vale only complain about some of the `camelCase` words?

prcr:
> LGTM. Would Vale only complain about some of the `camelCase` words?

Yes, I also found it strange that it only complained about one endpoint name but not the others. :thinking: I decided not to spend time investigating that. :sweat_smile:

PR #1618, opened 2023-02-07
fix: Fix incorrect sorting of self-hosted versions on a select menu DOCS-516

nicklem:
The ordering logic that sorts self-hosted versions on the version select dropdown breaks for numbers >= 10, causing the well-known and wrong sorting `1, 10, 2, 3, 4, ...`

image

### ❗ Note to reviewers

* In production, the dropdown is populated dynamically using data pulled from https://docs.codacy.com/versions.json. **This doesn't work when previewing,** since the `versions.json` endpoint returns 404.
* I've therefore [committed `versions.json` as test data](https://github.com/codacy/docs/pull/1618/commits/51e746319003e625dd27f4f517e5b67a5435c714) to populate the dropdown in the preview.
* This, in turn, causes some console errors as the JS tries to fetch and process the sitemaps of other versions (which are also not available on the preview). You can therefore ignore those console errors.

### :eyes: Live preview
https://fix-dropdown-menu-sorting-docs-516--docs-codacy.netlify.app/

### :construction: To do
- [x] Remove test data before merging.


docs/assets/javascripts/version-select.js
```diff
@@ -37,9 +39,51 @@ window.addEventListener("DOMContentLoaded", function() {
         var deprecatedVersionsArray = new Array();
         var versionsArray = new Array();
 
+        // Used to decorate options with a numeric sort order
+        function decorateWithSortOrder(option) {
```
inesgomas, line 43: I am very confused what is this all about 😱

docs/assets/javascripts/version-select.js
```diff
@@ -37,9 +39,51 @@ window.addEventListener("DOMContentLoaded", function() {
...
+            // Option version, formatted as major.minor.patch (e.g. 1.1.0)
+            var optionVersion = option.value.match(/[\d.]+/);
+            optionVersion = optionVersion ? optionVersion[0] : '';
+
+            // We split over an array for further processing.
+            var optionVersionArray = optionVersion.split('.').map(function (v) {
+                return parseInt(v, 10);
```
inesgomas, line 50: What does this do? 🤔

docs/assets/javascripts/version-select.js
nicklem: Here's the problem we're solving with this function. We're sorting versions that have dots, and this messes with the correct sorting when comparing strings:
```
> console.log('2.5.123' > '10.1.456')
> true // Whoops. Version 10 should come after version 2...
```

This function takes the version number of the option, say `2.5.123`, in the form `major.minor.patch` and transforms it into a number we can then use for sorting.

It does so by multiplying the main version number by 1 million, the subversion by 1 thousand, and the patch number by 1, and then summing them all.

So `2.5.123` becomes `2500123` and `10.1.456` becomes `10100456`, fixing our sorting issue:
```
> console.log(2500123 > 10100456)
> false // Better.
```

I know...lots of lines to solve a tiny problem. If you have better ideas, I'm all ears.
I've seen other solutions, even less readable than this one 😅

docs/assets/javascripts/version-select.js
nicklem: The same as `parseInt`. The second argument is the radix (in case you wanted to parse a hex string). I think the default is 10, so I could have omitted it.

But basically this code transforms `10.1.456` into `['10', '1', '456']` into `[10, 1, 456]`

docs/assets/javascripts/version-select.js
inesgomas: Yeah... I really can't read into this 🙈 I will approve to unblock you, but if you prefer another set of eyes, feel free to request their review

inesgomas (approved):

docs/assets/javascripts/version-select.js
```diff
@@ -1,3 +1,5 @@
...
..."version": "v4.0", "title": "Self-hosted v4.0", "aliases": []}, {"version": "v3.5", "title": "Self-hosted v3.5", "aliases": []}, {"version": "v3.4", "title": "Self-hosted v3.4", "aliases": []}, {"version": "v3.3", "title": "Self-hosted v3.3", "aliases": []}, {"version": "v3.2", "title": "Self-hosted v3.2", "aliases": []}, {"version": "v3.1", "title": "Self-hosted v3.1", "aliases": []}, {"version": "v3.0", "title": "Self-hosted v3.0", "aliases": []}, {"version": "v2.2", "title": "Self-hosted v2.2", "aliases": []}, {"version": "v2.1", "title": "Self-hosted v2.1", "aliases": []}, {"version": "v2.0.387-deprecated", "title": "Self-hosted v2.0.387 (Deprecated)", "aliases": []}, {"version": "v2.0", "title": "Self-hosted v2.0", "aliases": []}, {"version": "v1.5", "title": "Self-hosted v1.5", "aliases": []}, {"version": "v1.4", "title": "Self-hosted v1.4", "aliases": []}, {"version": "v1.3", "title": "Self-hosted v1.3", "aliases": []}, {"version": "v1.2", "title": "Self-hosted v1.2", "aliases": []}, {"version": "v1.1", "title": "Self-hosted v1.1", "aliases": []}, {"version": "v1.0", "title": "Self-hosted v1.0", "aliases": []}, {"version": "latest", "title": "Cloud (Latest)", "aliases": []}]
```
inesgomas, line 2:
```suggestion
```

we could clean this right?

docs/assets/javascripts/version-select.js
```diff
@@ -117,6 +155,8 @@ window.addEventListener("DOMContentLoaded", function() {
     }
 
     function generateVersionSwitcher(versionJSON) {
+        // TODO delete the following line
+        versionJSON = testData;
```
inesgomas, line 159:
```suggestion
```

docs/assets/javascripts/version-select.js
nicklem: Correct. This data is here just because it is not available on the preview. I hard-coded it to test locally and to make it available on this preview.

The reason: If you check another preview deployment related to some other PR, say [the CS2 one](https://feat-multiple-coding-standards-io-3--docs-codacy.netlify.app/), you'll see that the dropdown only has 1 version, which wouldn't allow us to confirm visually that the sorting is working properly.

PR #1629, opened 2023-02-23
fix: Update link to supported language strings

prcr:
The link was no longer pointing to the start of the array of language strings.


nicklem (approved):
Linking to a line in a repo...what could ever go wrong? 😆 

:shipit:

PR #1645, opened 2023-03-06
feat: Update submodule

nicklem:
Integrates https://github.com/codacy/codacy-mkdocs-material/pull/81


prcr (approved):
We can merge this change, but it only makes a difference in this repository if the actual compiled theme changes. :slightly_smiling_face:

PR #1646, opened 2023-03-06
toolchain: Run MkDocs when adding changes to a merge group

prcr:
This is necessary to validate that the docs site builds correctly and there are no broken links when using the [GitHub merge queue](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/merging-a-pull-request-with-a-merge-queue) feature.

See also [`merge_group`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#merge_group) for more information.

### DIFF
--- a/.github/workflows/mkdocs.yml
+++ b/.github/workflows/mkdocs.yml
@@ -1,6 +1,9 @@
 name: MkDocs
 
-on: push
+on:
+  push:
+  merge_group:
+    types: [checks_requested]
 
 # Set concurrency group to either "dev" or "prod"
 concurrency: ${{ fromJSON('{"false":"dev","true":"prod"}')[github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/release/v')] }}


prcr:
@nicklem I'd like us to try out the [GitHub merge queues](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/merging-a-pull-request-with-a-merge-queue) feature. :rocket: 

After this is merged, I'll activate the feature on the `master` branch protection rules.

nicklem (approved):
Let's try this!

prcr:
It seems like it's not possible to use this feature unless we do some more changes. After the pull request is added to the merge queue, all required CI/CD workflows must complete:

image

Although we can trigger all GitHub Action workflows with the event `merge_group`, I don't think there's a way to make Codacy analyze the merge queue branch as well:

image

The only option that I can think of would be to disable the mandatory Codacy status check.

PR #1658, opened 2023-03-22
feat: Create repository-level Coverage pages COV-256

prcr:
Adds repository-level pages specific to the Coverage product.

### :eyes: Live preview
- https://feat-create-coverage-pages-cov-256--docs-codacy.netlify.app/coverage/repositories/repository-dashboard/
- https://feat-create-coverage-pages-cov-256--docs-codacy.netlify.app/coverage/repositories/commits/
- https://feat-create-coverage-pages-cov-256--docs-codacy.netlify.app/coverage/repositories/files/
- https://feat-create-coverage-pages-cov-256--docs-codacy.netlify.app/coverage/repositories/pull-requests/

### :construction: To do
-   [x] Review and fix TODOs (I'll update the screenshots on a separate pull request)

PR #1685, opened 2023-04-18
clean: Reference commits on different branches

prcr:
While working on https://github.com/codacy/docs/pull/1658 I noticed that we have a reference to "issues on other branches", when in fact it makes more sense to say "commits on other branches".

### :eyes: Live preview
https://clean-branch-commits--docs-codacy.netlify.app/repositories/commits/


nicklem (approved):
Indeed! :shipit:

PR #1718, opened 2023-05-09
feat: Clarify that repositories can follow one CS at a time DOCS-551

nicklem:
Clarifies that repositories can only follow one coding standard at a time.

### :eyes: Live preview
https://feat-clarify-one-cs-per-repo--docs-codacy.netlify.app/organizations/using-coding-standards/

### DIFF
--- a/docs/organizations/using-coding-standards.md
+++ b/docs/organizations/using-coding-standards.md
@@ -6,6 +6,8 @@ Coding standards help you ensure that Codacy analyzes multiple repositories with
 
 When you customize the tools or code patterns of a repository that follows a coding standard, Codacy warns you that the repository will stop following the coding standard and asks for your confirmation.
 
+Each repository can only follow one coding standard at a time. Applying a new coding standard to a repository unassigns any previously applied coding standard.
+
 !!! important
     Coding standards turn tools with configuration files on and off. Those tool configuration files, however, take precedence over the code patterns defined on the coding standard.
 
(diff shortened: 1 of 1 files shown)


docs/organizations/using-coding-standards.md
```diff
@@ -6,6 +6,8 @@ Coding standards help you ensure that Codacy analyzes multiple repositories with
 
 When you customize the tools or code patterns of a repository that follows a coding standard, Codacy warns you that the repository will stop following the coding standard and asks for your confirmation.
 
+Each repository can only follow one coding standard at a time. Applying a new coding standard to a repository unassigns any previously applied coding standard.
```
nicklem, line 9: I'm not 100% sure about using "unassigns" as it's an unusual word, but I suspect that other variants (overwrites, replaces) may lead the user to believe that something destructive happens to the coding standard itself.

prcr: Sounds good! :ok_hand: 

The only thing that I'm on the fence about is whether this paragraph could switch positions with the previous one. So we'd talk about high-level information first, and mention the repository-level settings last. But that's just an idea. :bulb:

prcr: I saw that Vale complains about the word "unassigns", but this should be such a rare word that it's probably better to just ignore it locally:

```suggestion
Each repository can only follow one coding standard at a time. Applying a new coding standard to a repository <span class="skipvale">unassigns</span> any previously applied coding standard.
```

prcr (approved):

PR #1735, opened 2023-05-22
feat: Clarify scope of coding standard configurations DOCS-556

prcr:
- Clarifies the scope of the configurations that apply to a repository when assigning a coding standard
- Also refactors the "extra details" of how coding standards work under a single note admonition

### :eyes: Live preview
https://feat-coding-standard-tools-scope-do--docs-codacy.netlify.app/organizations/using-coding-standards/


prcr:
@stefanvacareanu7, these changes are meant to clarify your findings from [TS-367](https://codacy.atlassian.net/browse/TS-367) a bit better. Let us know what you think.

docs/organizations/using-coding-standards.md
```diff
@@ -4,24 +4,28 @@ Create coding standards on your organization to define and apply shared tool and
...
+## Creating a coding standard {: id="creating"}
 
 To create a coding standard for your organization:
 
 1.  Open your organization **Coding standards** page, tab **Coding standards**.
 
 1.  Click the button **Create new standard** at the top right-hand corner of the page. This opens a window with the coding standard creation form.
 
+    !!! note
+        Codacy currently supports up to 10 coding standards per organization.
```
nicklem, line 27: I like the idea of placing an admonition only when needed in the flow. Some of the ones we're grouping above could also be interspersed like this, for example `Each repository can only follow one coding standard...` under point 6. Thoughts?

nicklem (approved):
I left a comment for some optional edits, but LGTM otherwise :shipit:

docs/organizations/using-coding-standards.md
```diff
@@ -4,24 +4,28 @@ Create coding standards on your organization to define and apply shared tool and
 
 Coding standards help you ensure that Codacy analyzes multiple repositories with the same tool and code pattern configurations. For example, you can use a coding standard to ensure that a group of repositories follow the same security rules or coding conventions.
 
-Each repository can only follow one coding standard at a time. Applying a new coding standard to a repository <span class="skip-vale">unassigns</span> any previously applied coding standard.
+!!! note
```
claudiacarpinteiro, line 7: IMO, there is too much content in this admonition... 🤔 
We use an admonition to highlight a specific detail from the remaining content at a particular point. If we include a long piece of content, we miss the highlighting goal. 

An alternative would be to find a better spot in the flow to fit some of the bullets, as @nicklem suggested [here](https://github.com/codacy/docs/pull/1735/files#r1202081969).
Another one would be to keep the first three considerations, which are more descriptive/conceptual about coding standards, out of the admonition, and keep the admonition only for the last one, which refers to a specific precedence detail.
WDYT, @prcr ?

prcr: Indeed you're right, it seems that we've been adding lots of small details about how the coding standards work to this page and now the page is full of tips, notes, and important admonitions. :scream: 

I isolated the change of adding the new information in this pull request to make it self-contained, and will now open a different pull request for refactoring the information around the page in a tidier way.

docs/organizations/using-coding-standards.md
prcr: This is a good suggestion @nicklem, and I'll try to move information to where it's more relevant in a subsequent pull request (see https://github.com/codacy/docs/pull/1735#discussion_r1203871960).

PR #1740, opened 2023-05-25
feat: Clarify how to synchronize an org name DOCS-558

nicklem:
* Clarifies how to manually synchronize an organization name if it doesn't do so automatically.
* Rewords the existing sentence slightly to improve the flow.

### :eyes: Live preview
https://feat-org-name-sync-button-docs-558--docs-codacy.netlify.app/organizations/what-are-organizations/#updates-on-the-git-provider

### DIFF
--- a/docs/organizations/what-are-organizations.md
+++ b/docs/organizations/what-are-organizations.md
@@ -23,9 +23,12 @@ This opens the list of organizations on your Git providers. The organization wit
 
 ## Updates on the Git provider
 
-In case you change your organization or repository on the Git provider, some changes are automatically reflected on Codacy. The following cases are supported, depending on your Git provider:
+If you update your organization or repository information on your Git provider, some changes are automatically reflected on Codacy, as described in the table below.
 
-| Provider | Rename repository | Change repository visibility | Delete repository | Rename organization or group | Remove member from organization or group | Delete organization or group |
+!!! note
+    If an update to your organization name isn't automatically reflected on Codacy, navigate to the organization **Settings** page, tab **Profile**, and click the **Synchronize** button in the Synchronize organization name area.
+
+| Git provider | Rename repository | Change repository visibility | Delete repository | Rename organization or group | Remove member from organization or group | Delete organization or group |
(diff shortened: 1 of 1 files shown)


docs/organizations/what-are-organizations.md
```diff
@@ -23,9 +23,12 @@ This opens the list of organizations on your Git providers. The organization wit
 
 ## Updates on the Git provider
 
-In case you change your organization or repository on the Git provider, some changes are automatically reflected on Codacy. The following cases are supported, depending on your Git provider:
+If you update your organization or repository information on your Git provider, some changes are automatically reflected on Codacy, as described in the table below.
 
-| Provider | Rename repository | Change repository visibility | Delete repository | Rename organization or group | Remove member from organization or group | Delete organization or group |
+!!! note
+    If an update to your organization name isn't automatically reflected on Codacy, navigate to the organization **Settings** page, tab **Profile**, and click the **Synchronize** button in the Synchronize organization name area.
```
nicklem, line 29: I'm not a fan of admonitions, but this is an edge case and almost a troubleshooting instruction like the one in the previous section.

It would potentially work better as a footnote to the table's "Rename repository" heading, although I'm not sure we can do this with our setup @prcr.

prcr, line 29: Since the page only includes one **Synchronize** button and also because the page is so small, I suggest that we simplify this a bit:

```suggestion
    If an update to your organization name isn't automatically reflected on Codacy, navigate to the organization **Settings** page, tab **Profile**, and click the **Synchronize** button.
```

Otherwise, and for the sake of consistency, we should probably also make the name of the UI area bold:

```suggestion
    If an update to your organization name isn't automatically reflected on Codacy, navigate to the organization **Settings** page, tab **Profile**, and click the **Synchronize** button in the **Synchronize organization name** area.
```

prcr (approved):
This is great @nicklem, and if you publish this soon we'll be able to link the new SPA page [being prepared here](https://github.com/codacy/codacy-spa/pull/1194) to the updated docs section. :raised_hands:

PR #1754, opened 2023-06-01
feat: Mention that CI/CD needs access to Internet TS-395

prcr:
Running the Codacy Coverage Reporter [using our bash script](https://docs.codacy.com/coverage-reporter/alternative-ways-of-running-coverage-reporter/#bash-script) requires access to at least the following domains:

- `coverage.codacy.com` (downloading the bash script)
- `artifacts.codacy.com` (downloading the binary)
- `github.com` (downloading the checksum file)
- `api.codacy.com` or the domain of the Self-hosted instance (accessing the Codacy API)

As a simple approach to let users know about this, I suggest adding a note reminding users that their CI/CD platform needs to have access to the Internet generically, to avoid having the maintain the list of domains on the docs.

### :eyes: Live preview
https://feat-sh-access-script-ts-395--docs-codacy.netlify.app/coverage-reporter/alternative-ways-of-running-coverage-reporter/#bash-script

### DIFF
--- a/docs/coverage-reporter/alternative-ways-of-running-coverage-reporter.md
+++ b/docs/coverage-reporter/alternative-ways-of-running-coverage-reporter.md
@@ -26,7 +26,7 @@ The recommended way to run the Codacy Coverage Reporter is by using the [self-co
     ```
 
 !!! note
-    -   Make sure that your CI/CD platform has access to the Internet so that the command can download the bash script, the Codacy Coverage Reporter binary, and the validation checksum.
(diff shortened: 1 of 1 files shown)


claudiacarpinteiro:
@prcr If enterprise customers block access to the Internet from their CI/CD platform, maybe they want to know specifically to which domains Codacy Coverage Reporter requires access, so they can open the access only to those domains.
If we already have that information - which domains and the motive to access - we should consider providing this detailed information. WDYT? 🤔

prcr:
Thank you for your inputs @stefanvacareanu7. :bow: 

I'm going to close this pull request then, but we can always reassess if we need to add more information to the docs in case we understand that more customers have trouble with this. :+1:

PR #1760, opened 2023-06-06
feat: Explain how to override default ignored files DOCS-564

nicklem:
Explain how to override the default list of ignored files

### :eyes: Live preview
https://feat-override-default-ignored-files--docs-codacy.netlify.app/repositories-configure/ignoring-files/#default-ignored-files

### DIFF
--- a/docs/repositories-configure/ignoring-files.md
+++ b/docs/repositories-configure/ignoring-files.md
@@ -19,7 +19,7 @@ If you need more flexibility in ignoring files [use a Codacy configuration file]
 
 ## Default ignored files
 
-By default, Codacy ignores all files that match the following regular expressions:
+By default, Codacy ignores all the files that match the regular expressions below. If you want to override these defaults, [use a Codacy configuration file](codacy-configuration-file.md).
 
 ```text
 .*[\.-]min\.css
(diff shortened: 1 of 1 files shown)


docs/repositories-configure/ignoring-files.md
```diff
@@ -19,7 +19,7 @@ If you need more flexibility in ignoring files [use a Codacy configuration file]
 
 ## Default ignored files
 
-By default, Codacy ignores all files that match the following regular expressions:
+By default, Codacy ignores all the files that match the regular expressions below. If you want to override these defaults, [use a Codacy configuration file](codacy-configuration-file.md).
```
nicklem, line 22: Keeping it simple here. I'll just point the user to the relevant page like above.

prcr: We can probably give an extra hint on how users should use the Codacy configuration file while still keeping the sentence simple. Something along the lines of:

> If you want to override these defaults, [use a Codacy configuration file](codacy-configuration-file.md) to define a custom list of file paths to exclude.

This should be enough to lead customers into defining the `exclude_paths` array with their custom expressions.

prcr (approved):
:shipit: :+1:

PR #1776, opened 2023-07-02
toolchain: Automatically update Netlify pull request preview URL

prcr:
Automatically update the pull request body with the Netlify pull request preview URL.

### :construction: To do
-   [ ] If relevant, include the Jira issue key at the end of the pull request title
-   [ ] Perform a self-review of the changes
-   [ ] Fix any issues reported by the CI/CD


prcr:
I decided that this isn't a good solution because the pull request may not be created yet when the workflow runs (since it runs `on_push` and not `on_pull_request`).

PR #1785, opened 2023-07-07
fix: Rename Security Manager permission area

nicklem:
Fixes an oversight where we call the "Security managers" area in the UI using the old name, "Permissions".

### :eyes: Live preview
https://fix-rename-security-manager-permiss--docs-codacy.netlify.app/organizations/managing-security-and-risk/#assigning-the-security-manager-role


prcr (approved):
Indeed! Now it matches the copy on the screenshot and what's live in production. :rocket:

PR #1787, opened 2023-07-10
toolchain: Update GH actions to detect deleted and renamed files DOCS-535

nicklem:
Detects deleted Markdown files and posts a reminder to the PR thread.

TODO:
* [x] ❗ **Restore deleted/renamed files before merging**

### DIFF
--- a/.github/workflows/deleted-files.yml
+++ b/.github/workflows/deleted-files.yml
@@ -1,0 +1,40 @@
+name: Deleted files
+
+on:
+  pull_request:
+    branches:
+      master
+
+jobs:
+  list-deleted-files:
+    runs-on: ubuntu-latest
+    name: Detect deleted files
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Get changed docs md files
+        id: changed-files
+        uses: tj-actions/changed-files@v37
+        with:
+          files: docs/**/*.md
+
+      - name: Format list of deleted files
+        id: format-output
+        env:
+          ORIGINAL_DELETED: '${{ steps.changed-files.outputs.deleted_files }}'
+          ORIGINAL_RENAMED: '${{ steps.changed-files.outputs.renamed_files }}'
+        run: |
+          echo "FORMATTED_DELETED=$(echo $ORIGINAL_DELETED | sed 's/ /`<br>`/g')" >> $GITHUB_OUTPUT
+          echo "FORMATTED_RENAMED=$(echo $ORIGINAL_RENAMED | sed 's/ /`<br>`/g')" >> $GITHUB_OUTPUT
+
+      - name: Comment PR if there are deleted files
+        if: steps.changed-files.outputs.any_deleted == 'true'
+        uses: thollander/actions-comment-pull-request@v2
+        with:
+          message: |
+            ## Remember to add redirects
+            ${{ steps.format-output.outputs.FORMATTED_DELETED != '' && format('This PR **deletes** one or more `md` files:<br>`{0}`<br>', steps.format-output.outputs.FORMATTED_DELETED) || ''  }}
+            ${{ steps.format-output.outputs.FORMATTED_RENAMED != '' && format('This PR **renames** one or more `md` files:<br>`{0}`<br>', steps.format-output.outputs.FORMATTED_RENAMED) || ''  }}
+            Remember to append the appropriate redirects in `mkdocs.yml` under `plugins.redirects.redirect_maps`.
+          comment_tag: deleted-files-reminder


nicklem:
@prcr Ready for review and suggestions.

Some choices, like the format of the deleted files list, were forced by some limitations with GitHub's internal handling of newline characters.

.github/workflows/deleted-files.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: thollander/actions-comment-pull-request@v2
+        with:
+          message: |
+            ## Notice
+            
+            This PR deletes one or more `md` files:
+            `${{ steps.format-output.outputs.FORMATTED }}`
+            
+            Remember to append the appropriate redirects in `mkdocs.yml` under `plugins.redirects.redirect_maps`.
+          comment_tag: deleted-files-reminder
```
nicklem, line 37: FYI, this line causes the action to overwrite the comment, so we don't get spammed.

.github/workflows/deleted-files.yml
```diff
@@ -0,0 +1,38 @@
...
+        id: changed-files
+        uses: tj-actions/changed-files@v37
+        with:
+          files: docs/**/*.md
+      - name: Format list of deleted files
+        id: format-output
+        env:
+          ORIGINAL: '${{ steps.changed-files.outputs.deleted_files }}'
+        run: |
+          echo "FORMATTED=$(echo $ORIGINAL | sed 's/ /`<br>`/g')" >> $GITHUB_OUTPUT
```
nicklem, line 25: I tried different flavors of newlines here with little success, such as:

```suggestion
          echo "FORMATTED=$(echo $ORIGINAL | sed 's/ /\\n/g')" >> $GITHUB_OUTPUT
```

.github/workflows/deleted-files.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Comment PR if there are deleted files
+        if: steps.changed-files.outputs.any_deleted == 'true'
```
nicklem, line 27: I tried adding a similar action with a test on `'false'` to delete the comment if all `md` files are restored, but the behavior was odd, and it's an edge case, so ended up dropping it.

.github/workflows/deleted-files.yml
prcr: Try this instead:

```suggestion
          echo "FORMATTED=$(echo $ORIGINAL | sed 's/ /`  <br />`/g')" >> $GITHUB_OUTPUT
```

I'm thinking that GitHub's Markdown renderer appends the content of multiple lines together on the same "HTML line". According to the Markdown specification, we actually [need to add two spaces at the end of the lines](https://www.markdownguide.org/basic-syntax#line-breaks) to force them to be rendered with line breaks.

.github/workflows/deleted-files.yml
```diff
@@ -0,0 +1,38 @@
...
+      master
+
+jobs:
+  list-deleted-files:
+    runs-on: ubuntu-latest
+    name: Detect deleted files
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Get changed docs md files
```
prcr, line 15: Small style suggestion: maybe add an empty line between each step in the job. It helps with readability when checking out the workflow definition:

```suggestion
      - name: Checkout
        uses: actions/checkout@v3

      - name: Get changed docs md files
```

prcr:
@nicklem, I tweaked the title of the pull request. :arrow_up: :nerd_face:

.github/workflows/deleted-files.yml
```diff
@@ -0,0 +1,38 @@
...
+        id: changed-files
+        uses: tj-actions/changed-files@v37
+        with:
+          files: docs/**/*.md
+      - name: Format list of deleted files
+        id: format-output
+        env:
+          ORIGINAL: '${{ steps.changed-files.outputs.deleted_files }}'
```
prcr, line 23: As of now, the workflow is only reactive to **deleted** files. But it's actually more frequent for us to **rename** files, and there should also be a reminder for those situations.

I already checked and the GitHub Action also seems to output a separate list with the renamed files, so it should be a matter of perhaps appending the two lists together, or even outputting the two lists separately on the comment. Do you care to have a go at it? :nerd_face:

.github/workflows/deleted-files.yml
nicklem: Correct, there's no need to change this. My suggestion is only an example.
What I was trying to achieve was printing all files as a unique code block:
```
docs/faq/code-analysis/does-codacy-check-for-dependencies.md
docs/getting-started/adding-a-codacy-badge.md
docs/index.md
```

But this would not work, generating something like:
```
docs/faq/code-analysis/does-codacy-check-for-dependencies.md\ndocs/getting-started/adding-a-codacy-badge.md\ndocs/index.md
```

This is why I was trying a newline char instead of its HTML equivalent...but code blocks escape the newline char, too. Using any kind of literal newline would either break the pipeline or not display properly.

I consider this a cosmetic improvement though, and not worth banging our heads too much for.

prcr: Ahh, now I understand. But I agree, it's not worthwhile chasing this minor cosmetic improvement. :+1:

prcr (commented):
I like the new title for the comment and it seems like it's working great: https://github.com/codacy/docs/pull/1790#issuecomment-1631173805

Let's restore the deleted/renamed files and then merge. :shipit: :rocket:

nicklem:
I force-pushed to have a clean slate.
Ready for approval 👍

prcr (approved):
:robot: :tada:

PR #1767, opened 2023-06-21
feat: Release notes for Codacy Self-hosted 12.0.0 DOCS-561

lolgab:
Adds auto-generated release notes for Codacy Self-hosted 12.0.0

### :eyes: Live preview
https://feat-release-notes-self-hosted-12-0--docs-codacy.netlify.app/release-notes/self-hosted/self-hosted-v12.0.0/

### :construction: To do
- [x] Update the variable `extra.codacy_self_hosted_version` in `mkdocs.yml` to the new Codacy Self-hosted version
- [x] Add the new release notes page to `mkdocs.yml` and `docs/release-notes/index.md`
- [x] Add the breaking changes section if the release is a new major version (see [this Slite page](https://codacy.slite.com/app/docs/Fe1IpJXzBXRz_W))
- [x] Review generated release notes
- [x] Review list of issues with missing release notes
- [x] Ensure that links point to documentation for new Codacy Self-hosted version
- [x] Review links to changelogs of updated tools
- [x] Update the CLI tool versions and update the variable `extra.codacy_coverage_reporter_version` in `mkdocs.yml`
- [x] Update release date and remove `TODO` comments
- [x] Cherry-pick [this commit](https://github.com/codacy/chart/pull/788/commits/a88374c598af08da1f59e6f961371e0f14df6014) with recent v11 updates to `chart`'s [v12 branch](https://github.com/codacy/chart/tree/release-12.0.0) after merging the relevant PR: https://github.com/codacy/chart/pull/788
- [x] Update release notes (breaking changes?) to mention the removal of support for the custom worker scheduler in worker-manager (check with @lolgab), see [Slack message](https://codacy.slack.com/archives/CSSSHAK9N/p1687521426975769?thread_ts=1687166816.965619&cid=CSSSHAK9N)

PR #1789, opened 2023-07-11
fix: Update URLs for the Prospector and Pylint tools

prcr:
[While working on the Self-hosted release notes](https://github.com/codacy/docs/pull/1767#discussion_r1259711770) we noticed that the GitHub repositories for Prospector had moved to a different organization. While searching for `PyCQA` I noticed that the URLs for Pylint also had redirects. This pull request updates all relevant URLs to the newest repositories for the tools.

### DIFF
--- a/docs/getting-started/supported-languages-and-tools.md
+++ b/docs/getting-started/supported-languages-and-tools.md
@@ -261,8 +261,8 @@ The table below lists all languages and frameworks that Codacy supports and the 
       <td>Python
       </td>
       <td><a href="https://github.com/PyCQA/bandit">Bandit</a>,
-          <a href="https://github.com/PyCQA/prospector">Prospector</a>,
-          <a href="https://pylint.pycqa.org/">Pylint</a></td>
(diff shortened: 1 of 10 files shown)


nicklem (approved):
LGTM :shipit:

nicklem:
@prcr Should we update the link checker to fail on redirects? There's [an option for that](https://github.com/lycheeverse/lychee/blob/2231ccc618e0a06b8766bee9aa1353c9f20958ab/README.md?plain=1#L295-L296).

prcr:
> @prcr Should we update the link checker to fail on redirects? There's [an option for that](https://github.com/lycheeverse/lychee/blob/2231ccc618e0a06b8766bee9aa1353c9f20958ab/README.md?plain=1#L295-L296).

So this explains why we weren't getting notified in this situation!

Before, we were using a different tool that would complain of redirects by default (in fact, it detected so many false positives that I had to find lychee as an alternative).

So yes, it's better to activate redirect detection so we're not eventually surprised by broken links. Care to do it?

PR #1809, opened 2023-07-28
feat: Reorder SRM page sections

nicklem:
A quick fix to move the section "Managing access to security and risk management" of the SRM page further up and reestablish the concept/task/reference structure of the page.

### :eyes: Live preview
[Managing access to security and risk management](https://feat-reorder-srm-sections--docs-codacy.netlify.app/organizations/managing-security-and-risk/#managing-access-to-security-and-risk-management)

### DIFF
--- a/docs/organizations/managing-security-and-risk.md
+++ b/docs/organizations/managing-security-and-risk.md
@@ -96,6 +96,40 @@ The following section explains how to integrate with each of the supported platf
 
     1.  Find the Jira entry under the Integrations list and click the **Options** button (identified by three dots), then select **Uninstall** and confirm.
 
+## Managing access to security and risk management
+
+To better track and address security issues, organization admins can extend access to security and risk management by assigning the Security Manager role to organization members.
+
+The Security Manager role is independent of the Git provider role of an organization member and provides only the permissions necessary to monitor security issues, following the principle of least privilege:
+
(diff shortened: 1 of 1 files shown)


docs/organizations/managing-security-and-risk.md
```diff
@@ -131,6 +131,40 @@ The following section details when Codacy opens and closes items for each suppor
         -   If an issue is opened and closed on the same day, Codacy may not detect it.
         -   To make sure that Codacy detects Jira issues correctly, assign the "security" label when creating the issue or immediately after.
 
+## Managing access to security and risk management
```
claudiacarpinteiro, line 134: Does it make sense to move the **Managing access to security and risk management** section right after the **Managing integrations** instead? 🤔 

This way we would gather the two procedures that users can do on the screen, and keep reference information below the procedures.

Also, the **Opening and closing items** section looks like reference information, because they're not actions that users can perform. A couple of notes on this:
-   We are naming procedure sections with a gerund, such as Managing integrations, Installing/Uninstalling Jira integration, and so on. Thus, maybe we should rename this section to something not gerund, to avoid the expectation of a user procedure. Eg, **Open and close items**.
-   Is the content in **Open and close items** related to "Item statuses"? Does it make sense to gather these two in the same section? 🤔 

Thoughts?

claudiacarpinteiro (approved):
🚀

PR #1957, opened 2023-11-23
fix: clarify how auto-follow works

mrfyda:
Clarify that the Follow is automatic only for Organization admins, not Organization managers, members, or anyone with Repository Admin permissions.

also, disregard who adds the repository: the auto follow happens always. when an organization admin adds it, or via API, or by repository admins.


machadoit (approved):

jorgebraz (approved):

nicklem (approved):
LGTM 👍 As long as it's correct, it's clear and concise.

PR #1968, opened 2023-12-04
feat: Update the use of "framework" DOCS-615

nicklem:
Update the use of the term "framework" to prevent confusion when browsing the relevant pages.

### :construction: To do
- [x] Confirm if we should explicitly list frameworks in any measure. We may decline this due to maintenance overhead.
  - Decision: declined for now. let's revisit this in the future if more requests emerge in this sense. It's a sizable effort.


mrfyda:
LGTM 👍

PR #1980, opened 2023-12-15
feat: Clarify upload example limitations DOCS-589

nicklem:
Clarifies that the user should use GNU find to run an example.

### :eyes: Live preview
https://docs-589-update-example-on-uploadin--docs-codacy.netlify.app/coverage-reporter/uploading-coverage-in-advanced-scenarios/#multiple-reports-once


docs/coverage-reporter/uploading-coverage-in-advanced-scenarios.md
```diff
@@ -31,6 +31,9 @@ bash <(curl -Ls https://coverage.codacy.com/get.sh) report \
     -l Java $(find . -name 'jacoco*.xml' -printf '-r %p ')
 ```
 
+!!! note
+    The example above works only on systems that use GNU find with support for the `-printf` action, such as Linux.
+
```
nicklem, line 36: @claudiacarpinteiro Should we place this admonition above the example? It's not as clean but the user will definitely notice it. Having it below the example may cause the user to ignore it. Thoughts?

claudiacarpinteiro: I think it would work fine above the example, which is usually the way you prefer, right? 😉 

Also, we could format the "find" as code (the suggestion below is worded for the admonition before the example).

```suggestion
!!! note
    This example works only on systems that use GNU `find` with support for the `-printf` action, such as Linux.

```

claudiacarpinteiro (approved):
I left a small comment, but it's already good to go! 👍

PR #2077, opened 2024-04-12
feat: Remove docs from the default ignored files DOCS-663

claudiacarpinteiro:
Removes the docs folder from the list of default ignored files.

### :eyes: Live preview
https://docs-663-remove-docs-from-ignored--docs-codacy.netlify.app/repositories-configure/ignoring-files/#default-ignored-files


nicklem (approved):
👍

PR #2126, opened 2024-06-04
feat: Fix warning on edge case when building docs

nicklem:
Fixes a weird warning that prevents from building docs locally (unless forced to ignore warnings).

This looks like an mkdocs bug, where an anchor link imported in a fragment imported from the same directory is rendered with an extra dot in front: `<a href=".#possible-issues">potential new issues, or potential fixed issues</a>`


nicklem:
Turns out it was a Python version issue: we should use Python 3.9 and not 3.10 (at least for now)/

Closing this one.

PR #2152, opened 2024-06-26
fix: Codacy config file must start with `---`

mrfyda:
This is not required for the Automatic summary feature but existing Codacy processing will fail.

> The file must start with a line containing a triple dash (---).

It's best to follow the recommendation and avoid that users creating their file for Automatic summaries are then faced with this problem later on.


nicklem (approved):
👍

PR #2686, opened 2026-06-16
Add 'Dependency chain' heading to security findings page

mark-rln:
Gives the dependency chain paragraph on the **Managing security and risk** page its own `### Dependency chain` subheading (under the **Findings** section), so it can be linked to directly and is easier to scan.

No content changes — only the heading was added.


andrzej-janczak (approved):
LGTM

